PR #12, opened 2020-06-29
Changed terminology for pending transactions

Joohansson:
Replaced "Pending" with "Incoming".
Replaced "Pending Deposits with "Incoming Amount"


BitDesert (approved):
Looks fine to me for a first test.
I would like to see some action in this space, if it does not fit the terminology everyone aggrees on we can change it again.

PR #15, opened 2020-06-29
Add support for expanded private keys

PlasmaPower:
Also exposes support for normal private keys. I've copied this PR from https://github.com/cronoh/nanovault/pull/93

Explanation of expanded private keys: Basically, private keys are first hashed, and that hash is the "expanded private key" that the signing elliptic cryptography is done on. A lot of fancy cryptography like "public seeds"/non-hardened derivation, where you can generate addresses from a public seed without knowing corresponding private keys from the private seed, is built using elliptic cryptography, so the keys it generates are the elliptic cryptography input to the signing algorithm, aka the expanded private key.

### DIFF
--- a/src/app/components/configure-wallet/configure-wallet.component.html
+++ b/src/app/components/configure-wallet/configure-wallet.component.html
@@ -99,6 +99,22 @@
         </div>
       </div>
     </div>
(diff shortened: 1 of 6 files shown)


BitDesert (approved):

Joohansson (approved):
Tested importing a key and seems to work fine. I'm good with this

PR #16, opened 2020-06-30
Random backend

### LINKED ISSUE #8: Automatic Backend switching
Currently most backends are rate limited.
For UX and decentralization purposes it's better to rotate the backend at first start, when a backend goes offline or you run out of tokens.

BitDesert:
- uses a random backend until you set in in the settings
- it switches to a different one if you encounter a rate limit (but not when it's offline, but that can be changed)


BitDesert:
This solves #8

Joohansson (commented):
I think the default websocket for nanex and nanocrawler needs to fixed first (for example set it to ninja or nanos.cc) or users will randomly not be able to auto-receive. I'm also a bit worried that nanex work_generate will be "randomly slow" for users. Not sure how to handle that.

Joohansson:
Looks good to me now!

PR #21, opened 2020-06-30
Update Angular to 9.1.11

### LINKED ISSUE #9: Update Angular Version
The used Angular version is 5 and is not supported anymore.
It would be better to update it to a more recent version.
The process could break features!

BitDesert:
This solves #9 and updates the Angular version to 9.1.11.
Sending/Receiving works but needs further testing.


package.json
```diff
@@ -30,46 +30,48 @@
...
     "crypto-js": "^3.1.9-1",
     "electron-updater": "^2.21.0",
     "hw-app-nano": "^1.3.0",
-    "ngx-clipboard": "^9.1.2",
+    "ngx-clipboard": "^12.3.0",
     "node-hid": "^1.3.0",
     "qrcode": "^1.4.4",
-    "rxjs": "^5.5.2",
+    "rxjs": "^6.5.5",
+    "rxjs-compat": "^6.0.0-rc.0",
```
NiFNi: @BitDesert is there a specific reason to go to a release candidate that is 2 years old? Seems a bit wrong to not just use the current version here. Even if npm uses it if I understand it correctly.

src/app/components/configure-wallet/configure-wallet.component.ts
```diff
@@ -202,7 +202,7 @@ export class ConfigureWalletComponent implements OnInit {
     const file = files[0];
     const reader = new FileReader();
     reader.onload = (event) => {
-      const fileData = event.target['result'];
+      const fileData = event.target['result'] as string;
```
NiFNi: Yeah I was stuck here aswell. I am unsure if we need to handle the case that we get an ArrayBuffer instead of a string here. Are you sure that it will not create any problems?

NiFNi (commented):
@BitDesert apart from those two comments it's pretty similar to what I did.

package.json
BitDesert: Hmm didn't change that by hand, probably a default from the Angular update process.
We can push that easily to a more current version.

NiFNi: Yeah I think going to latest should be fine. Major version is still 6.

BitDesert: Updated to 6.5.5

Joohansson (commented):
Ledger nano S does not work with the desktop app on windows. It does however work on the web, also on the master branch desktop app. Something has happened!

src/app/components/configure-wallet/configure-wallet.component.ts
BitDesert: In which case would we get an ArrayBuffer here? We call `readAsText` so it should always return a string.

NiFNi: Ah I see how this works. So yeah it should always be a string.

Joohansson:
Added some troubleshooting steps. The ledger does not load on desktop version and funny enough it has nothing to do with the ledger libraries.

I know the exact line of cause and it's here: https://github.com/BitDesert/Nault/blob/5be2d36902f582146dd485326b2c8c615073e009/src/app/services/desktop.service.ts#L13

It's a pure electron IPC issue where the window object is undefined when it was not in earlier angular/electron version. Spent two days troubleshooting this and now I give up. It should be undefined when running on web, but not on desktop.

It's related to this guide: https://dev.to/michaeljota/integrating-an-angular-cli-application-with-electron---the-ipc-4m18
And by reading the comments we see others have the problem too but I have also seen people manage to fix it, so somehow it should be possible to solve. It's just complicated.

**Related section from the guide:**

> The history of require: require as been with us for a while, since the first version of Node back in 2008, almost 10 years. But still is one of the most misunderstand functions in modern web development. With the integration of import and export keywords in Node, several articles was written to explain how the require function currently works. TL;DR: A require function is injected for every file, allowing Node to resolve the dependencies. Later, module builders will look for import and require and will try to resolve modules, assuming that's what you want.
> 
> So, now that we know that require is actually a function injected by Node, then how is require able to work in Electron renderer process. You may guessed it, Electron injects its own version of require function in the global scope of the renderer process when it loads the page. So, although it may seems like the same function, it is not.

> As you can see, we will first check if window object has a require property. With this we will assume we are inside Electron, then we will try to require('electron'), if for any reason it doesn't work it just throw an error, and the property _ipc will be undefined. Checking require in the window object will allow us to run the service in a regular browser context, in that case the _ipc won't have a assignment value and will be undefined.

The troubleshooting step to solve this can be done in the desktop app dev tools (ctrl+shift+I) where the first console line is
`Electron's IPC was not loaded. Normal on web but not desktop.`

That's happening when the "window" is undefined. No ledger device is required to solve this issue. When that line is gone, the Ledger should work normally!

I also found references to a security variable that was changed in electron 5 that should have solved this. I added
`nodeIntegration: true`
to the electron webPreferences, but no difference observed. Files changed:
https://github.com/BitDesert/Nault/blob/257a10b2b0d4caf958625f4d9b2c0deaa0763f8e/main.js#L26
https://github.com/BitDesert/Nault/blob/257a10b2b0d4caf958625f4d9b2c0deaa0763f8e/desktop-app/src/desktop-app.ts#L18

Also found this that didn't help me at all. Got other issues instead with the "normal" require in other functions.
https://www.reddit.com/r/reactjs/comments/dni84p/windowrequireelectron_is_undefined_when_app_is/

Joohansson:
Desktop app works with ledger on linux. It's just Windows 10 that has the problem above. Maybe some kind of security conflict?

Edit: Eeeeh now ledger works on desktop in windows too after restarting the computer and recompiling. I'm speechless.

Issue #10, opened 2020-06-29
Optional auth headers for the custom backend API


Joohansson:
Could be useful if you run a node on a remote server and want to point Nault to it over https and with authentication required. For example, using [NanoRPCProxy](https://github.com/joohansson/nanorpcproxy) as middleware.

Reference: https://tixwallet.cc/#/node

PR #24, opened 2020-06-30
Support for custom API auth headers

Joohansson:
Could you take a look at this @BitDesert if you think the GUI is ok?

I have verified that the authentication works against one of my public APIs that has it enforced. It will typically always be empty for predefined APIs since creds can't be perfectly hidden anyway in a client-side wallet.

Ref: #10


BitDesert:
Works with https://mynano.ninja/api/node 👍

BitDesert (approved):

Issue #23, opened 2020-06-30
Receive incoming transactions by amount: largest first


Joohansson:
As some API endpoints have rate limiting I think it makes more sense to receive pending sorted by amount instead of timestamp. That would solve the issue for accounts with hundreds of incoming that would like to save the small amounts for later. Or if the PoW takes a long time (it will be 8x slower later on).

PR #33, opened 2020-07-01
Setting for incoming tx with largest amount first

Joohansson:
Affects both incoming transactions for the account history view and the order of how they are processed. Regardless of the wallet was closed or opened while they arrived. Sending a large tx while smaller ones are processed will get the largest first as well.

Solves #23 

Available as app setting with Yes as default value.

Question is if the list should be sorted or not but I think it's an advantage if 1000 smaller ones are there and you are just interested in the large ones.

**EDIT:
The list is now unaffected by the setting. The min receive amount can be set to filter out spam.**

### DIFF
--- a/src/app/components/configure-app/configure-app.component.ts
+++ b/src/app/components/configure-app/configure-app.component.ts
@@ -94,8 +94,8 @@ export class ConfigureAppComponent implements OnInit {
   selectedPoWOption = this.powOptions[0].value;
 
   pendingOptions = [
-    { name: 'Largest Amount First', value: 1 },
-    { name: 'Oldest Transaction First', value: 0 },
+    { name: 'Largest Amount First', value: 'amount' },
+    { name: 'Oldest Transaction First', value: 'date' },
   ];
   selectedPendingOption = this.pendingOptions[0].value;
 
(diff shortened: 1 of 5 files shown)


aleksre:
Nice feature :)

Only thing I'm thinking is it might not be clear to users what the 'no' option means. Maybe if the input label was something like "Receive priority" the options could be "Largest first" or "Chronological"?

Joohansson:
> "Receive priority" the options could be "Largest first" or "Chronological"?

@aleksre Good suggestion. Now implemented as "Largest Amount First and "Oldest Transaction First"

Also, the list is now unaffected by the setting. The min receive amount can be set to filter out spam.

src/app/components/configure-app/configure-app.component.ts
```diff
@@ -93,6 +93,12 @@ export class ConfigureAppComponent implements OnInit {
   ];
   selectedPoWOption = this.powOptions[0].value;
 
+  pendingOptions = [
+    { name: 'Largest Amount First', value: 1 },
```
BitDesert: I would use strings as the `value`, they are easier identifiable.

BitDesert (approved):
Fixed by using strings

PR #35, opened 2020-07-04
First steps for i18n

### LINKED ISSUE #18: Internationalization support
I think it would be great to have support for multiple languages. It lowers the barrier to join the Nano world for non fluent english speakers.

Angular ships with i18n support. More information can be found here: https://angular.io/guide/i18n

If noone started on this already I think I will probably sooner or later start working on this.

NiFNi:
Quick explanation on how the angular i18n works:
Basically you just add a i18n attribute to a tag and explain the content in the tag using this structure: `<meaning>|<description>`
If the actual text is in another attribute like for example the title you can just add a `i18n-title` attribute.
Using `ng xi18n` you then generate the messages.xlf. To create a new translation we basically just take that file and add a target tag for each source tag in that file. In the target the translation for the source should be in. the meaning and description is also found in the `messages.xlf`. 
As an example I added the attributes to the app.component.html. Also I created the german translation for it which can be found in the messages.de.xlf. 
Currently the build process is not changed and should produce basically the same app as before this PR. If you want to test/see the german translation you have to run `ng build --localize`. That will create a build where localization is turned on. In that case for every locale setup in the angular.json a subfolder is created in the dist folder. Each contains the localized app. 

For now I think it would be great if this goes to master already and no new texts are added without a proper i18n configuration. I would continue on refactoring the existing texts component by component and regularly create PRs to keep possible merge conflicts as low as possible. As I said unless turned on the behaviour of the app should be exactly as before applying these changes.

relevant for #18


BitDesert:
Thanks for your PR, looks good! 
How is the switching done and how is the routing on the application root `/` done?
Ideally the app would use the language from the browser automatically.

NiFNi:
@BitDesert Angular seems a bit bad in that regard. But the way to go is probably just an index.html with plain js which redirects to the correct locale (loaded from the localstorage or if not available from the browser settings). Later we could add a setting to the "App Settings" which can overwrite the default given by the browser and saves to localstorage.

Additionally all subpaths of a locale have to be redirected by the server where the app is deployed. So for example /en/send has to redirect to /en/index.html. Example for nginx: https://stackoverflow.com/questions/49889093/nginx-configuration-for-angular-i18n-application

Joohansson:
Nice, yeah it must be in the settings. Nothing is more annoying than sites trying to translate to my language when I just want it in English (and don't allow me to change or hides it real hard).

NiFNi:
I got it running with a setting for the language. Some small things need a bit more work though. I'm done for today working on this so I will most likely push it tomorrow or the day after.

NiFNi:
Okay, I got most of it running. Some small things that probably need some polishing. To get an idea on this works and looks like during deployment: http://nault.nifni.net

The nginx config looks like this:
```
server {

    listen 80;

    server_name nault.nifni.net;

    location / {
        root /var/www/nault/;
        index index.html;
    }
    location /en/ {
        alias /var/www/nault/en/;
        try_files $uri$args $uri$args/ /en/index.html;
    }
    location /de/ {
        alias /var/www/nault/de/;
        try_files $uri$args $uri$args/ /de/index.html;
    }
}
```

Not sure how the deployment with vercel works. Is it something that is configured in the repo or is it some settings directly in vercel?

EDIT: Got it running on vercel with my repo. You can see the result here: https://nault-two.vercel.app 
I'll push that to this branch later today.

BitDesert:
Works with that little fix I added.
Should we use something like https://github.com/ngx-translate/core ? Could be better and more flexible than compiling the whole project for each language. But I don't know if that's the way to go...  It seems that `@angular/localize` is the go-to method from Angular >9.
Any feedback?

NiFNi:
> Works with that little fix I added.
> Should we use something like https://github.com/ngx-translate/core ? Could be better and more flexible than compiling the whole project for each language. But I don't know if that's the way to go... It seems that `@angular/localize` is the go-to method from Angular >9.
> Any feedback?

My vote is @angular/localize without further complexity. I believe that this will get a bit more love by the angular team in the future. Changing the language is something that is done so irregularly in normal usage (normally just once in the beginning) that the necessary reload does not really cause problems IMHO.

aleksre:
How is the tooling for managing translations using @angular/localize? I have not used it, but have very good experience with going the JSON route (like ngx-translate) on other projects which seem to be the standard for frameworks like Vue, React and Svelte. For example, [i18n-ally](https://github.com/antfu/i18n-ally) is an amazing VS code plugin that makes it easy to keep track of which languages have missing texts etc.

NiFNi:
@aleksre the files follow the XLIFF standard. After a quick look into the i18n-ally plugin the example for angular also uses the xliff files. So it seems to be supported. But I would actually suggest to create a project in [POEditor](https://poeditor.com/) or [Crowdin](https://crowdin.com/). Both offer free unlimited plans for open source projects. This would enable non tech-savvy people to also contribute to translations.

NiFNi:
I didn't get i18n running in the desktop app yet. Not sure if it is because of my incompetence with everything that involves the JS world or if there are actually real world problems :D If anyone wants to have a look at that feel free to try. 

Anyway I would suggest to already merge this without the desktop app supporting it. Unless changed in the angular.json desktop build configuration it should just build to english by default without i18n. 
If this is merged we can require to include i18n information for all new texts. And the merge conflicts will probably not get better if we wait longer with this and more other PRs come in :P

BitDesert:
Yep, I would merge it to master ASAP when we have a plan for the desktop wallet. With that redirecting index.html it should be possible, but I have to try that out.

PR #36, opened 2020-07-05
Improved receive screen

Joohansson:
The old receive screen is lacking use case. It's always hidden in the menu accept when the wallet is locked and new incoming tx has been detected. Then you can see them in the list but to receive you must unlock but at that time the wallet has already auto-received them. There is only one use case of that screen; if using Ledger and it messes up the receive block and you have to do it again manually.

The old screen looks like this:

**The changes coming with this PR:**

- Always visible in the menu, with a "new" flag visible if there are new incoming
- Same functionality as before to check incoming and do manual receive but when you select an account:
- Display QR (standard syntax for wallets to pick up deeplink)
- Optional amount to be encoded in the QR
- A function that checks if an input string is a valid Nano amount (numerical, not too little, not too big). Also implemented at the "min receive" in the settings page.

This is more in par with how other wallets work and the user should feel more familiar than having to go to the account page, click on the account and then get a QR than is not even compliant with deep link (nano:nano_abc).

New screen:

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -67,14 +67,16 @@
           <li routerLink="/accounts" routerLinkActive="uk-active"><a routerLink="/accounts" routerLinkActive="active" class="uk-margin-left">Accounts</a></li>
           <li><a routerLink="/send" routerLinkActive="active" class="uk-margin-left">Send</a></li>
           <li>
-            <div uk-grid>
-              <div class="uk-width-3-4">
-                <a routerLink="/receive" routerLinkActive="active" class="uk-margin-left">Receive</a>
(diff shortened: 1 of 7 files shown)


Joohansson:
Thanks! Looks super smooth on my mobile and no graphic glitches on collapse as the original.
There may be some minor conflict with #36 but probably only for the mobile part you deleted. That should be easy to solve.
I'm ok with this PR.

aleksre:
Looks great 👍 
Tested QR code with Natrium

BitDesert (approved):
Fixed the sidebar height (was not uniform).

PR #39, opened 2020-07-05
Consolidated nav menus into a single instance

aleksre:
This PR removes the mobile-only nav menu markup which is about 100 lines of duplicated code. All screen sizes now use the same code and nav instance without altering the UX.

**The PR also fixes some minor issues related to the mobile nav:**
- A bug reported on reddit where both nav instances would be visible at the same time: https://www.reddit.com/r/nanocurrency/comments/hhxjfv/the_rebirth_of_nanovault_nault/fwdwt91?utm_source=share&utm_medium=web2x
- Clicking "Representatives" or "Nault" (top left) would not close the nav drawer.
- Expand/collapse animation, which relied on UI kit off-canvas, was not completely smooth. New implementation is using pure CSS instead of off-canvas.

### DIFF
--- a/src/app/app.component.css
+++ b/src/app/app.component.css
@@ -88,6 +88,9 @@
   font-size: 10px;
 }
 
+.content-container {
+  z-index: 1;
+}
 
 @media (max-width: 12450px) {
   .content-container {
@@ -99,5 +102,18 @@
   .content-container {
     padding-left: 30px;
     padding-right: inherit;
+    transition: transform 0.5s ease;
+  }
+
+  .content-container.nav-expanded {
+    transform: translateX(350px);
+  }
+}
+
+@media (max-width: 639px) {
+  .nav-container {
+    position: absolute;
+    width: 350px;
+    height: 100%;
   }
 }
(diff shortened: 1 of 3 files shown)


Joohansson:
Thanks! Looks super smooth on my mobile and no graphic glitches on collapse as the original.
There may be some minor conflict with #36 but probably only for the mobile part you deleted. That should be easy to solve.
I'm ok with this PR.

Joohansson:
Out of curiosity, what does this do @aleksre ? https://github.com/BitDesert/Nault/pull/39/files#diff-5c26d2c8f8838f32ace87fdecd6344eaR49

aleksre:
> Out of curiosity, what does this do @aleksre ? https://github.com/BitDesert/Nault/pull/39/files#diff-5c26d2c8f8838f32ace87fdecd6344eaR49

It will set the expanded flag to false whenever a route change on the angular router is triggered. This will in turn close the drawer if it is open, which means that we no longer have to add toggle handlers to each link individually (this was missing from the Representatives and the Nault title links). It should be handled automatically for all current and future nav links now, assuming the link also change the route.

PR #41, opened 2020-07-06
QR Code support

BitDesert:
QR code scanning from inside the app.
Supports sending + Representative change


src/app/components/qr-scan/qr-scan.component.html
```diff
@@ -0,0 +1,58 @@
...
+        </div>
+
+      </div>
+
+    </div>
+
+  </div>
+</div>
+
+<div *ngIf="hasPermission === undefined" class="uk-alert uk-alert-info">
```
aleksre: According to [this chart](https://github.com/zxing-js/ngx-scanner/wiki/%60hasPermission%60-and-%60hasDevices%60-cheat-sheet), `hasPermission` and `hasDevices` could also be `null`, so I would suggest we check for `hasPermission == null` which will cover both the null and undefined cases.

Might also be a good idea to add this the type information to the .ts file when these variables are initialized. For example (line 27): `hasDevices: boolean|null = null;`

Joohansson:
Seems to work fine with embedded amount too. Nice work! @BitDesert You think we can make it detect seed and mnemonic too and auto-paste that into the sweeper source wallet? Or maybe you want to scan a paper wallet and use it to create a new wallet. Or maybe ask the user where to use it. Natrium can do both. What do you think?

BitDesert:
@Joohansson When you scan a seed you should be redirected to the sweeper if you already have a wallet or to the wallet import if you don't have a wallet set up.

Joohansson:
There is also a case when you don't have a wallet but still want to scan and use the sweeper tool but your suggestion is good enough. We want people to use the wallet after all :)

src/app/components/qr-scan/qr-scan.component.html
BitDesert: At the bottom of the doc there is a table with the states:
grafik

So `undefined` is the initial state and get's updated to `null` if there is a weird error. So we should leave the initial state as it is, but I updated the test cases for the alerts. That fixes the weird bug of having two alerts at once. Thanks for the review!

Joohansson (approved):
Verified in win10 chrome and iphone using safari. Chrome on iPhone won't work with cam but that's just the way it is. win10 desktop app doesn't work either because of some security check. Maybe not possible to solve?

BitDesert:
Now you can import seeds as well :)
Prepared for the sweeper, so one route is missing until it's merged.

Issue #29, opened 2020-07-01
Add possibility to send all funds from a wallet


MikeRow:
Add possibility to send all funds from a wallet, so you don't have to send manually from every account

Joohansson:
Interesting idea! So like empty a wallet. Only problem I see with that is if the user has not added all accounts containing funds and thinking they have emptied the wallet when they have not. There are billions of accounts in a seed, can't check for funds in all of them.

MikeRow:
Perhaps an automatic account search could discover accounts with funds on them. 
Taking (again) Electrum as example, Nault could look up to 10 next accounts, of the same seed.
Just looking, but not generating in the wallet, and notifies the user if there are some available funds.

Joohansson:
Yeah sure, it can check for hidden funds, no problem but each check require one API call and the public APIs have rate limiting. 10-20 is ok but there could still be rare cases people have more. I think this would be a nice feature. Could add a warning that there is no guarantee that all accounts are checked or something.

Joohansson:
Actually there is a tool that does this. I wrote it myself a while back. You just enter a seed, private key or mnemonic and it will send all funds to another account. It support any index but max 100 each time. It's using local pow only so a little bit slower than dpow. It will also receive any pending that may be waiting. It even support Ledger passphrase in case anyone need to recover that. 

https://tools.nanos.cc/?tool=sweep

Joohansson:
@mikerow Nault now has a wallet sweeper that can be used on its own seed if you want to empty the wallet

PR #42, opened 2020-07-06
Implemented wallet sweeper

### LINKED ISSUE #38: Implement a wallet sweeper
This is related to #29 but a correctly built sweeper can not only import any funds from any seed to deposit into the Nault wallet, but also if using your current Nault wallet as source seed and an external account it could also empty the whole wallet with one single click. Double functionality which I don't think any other wallet has.

I want to build this if no one tells me not to. Will probably allow any custom range of index to check (with limitation) similar to this tool: https://tools.nanos.cc/?tool=sweep

Joohansson:
I would like some help battle testing this @BitDesert . The core processing code is pretty much the same as the one I built for https://tools.nanos.cc/?tool=sweep but using the features of Nault, like the PoW source and minimum receive from the settings. It probably needs some design love but it's working.

An easy method for testing this is to generate a seed here: https://tools.nanos.cc/?tool=seed
Then using the QR on the account field and try sending some nano to different indexes and derivation method from Natrium. Then testing with either seed, private key or mnemonic on the sweeper. This is also a perfect simulation of a ledger passphrase recover if using the BIP39/44.

Solves #38 and #29


BitDesert (changes requested):
Some things I would change:
- Embed the GUI in a `uk-card` like the send dialog
- Remove the "Destination Account field" as you would use it only with accounts setup in Nault
- "Set Min" is basically a "reset" button, so we might call it that way
- "Set Max" sets currently both Start and End pretty high but 100 apart. Why?
- There is a third "Set Max" button that looks the same but just sets "Max TX" back to 100 (which is the default). So it's functionality is the same as a "Reset" button
- The message "An account was swept and funds transferred to the destination!" does not disappear, maybe include the account name as well
- A spinning animation like a the "Send" page would be cool, so you see that it's "doing something"
- There is no "Success" message when it's done

Would it be better to set the type of the secret in a dropdown? Or just let it fiddle out itself?

Joohansson:
New design
image

src/app/components/sweeper/sweeper.component.html
```diff
@@ -0,0 +1,121 @@
...
+            </div>
+          </div>
+
+          <div class="uk-width-1-1 narrow-div">
+            <div class="uk-form-horizontal">
+              <div class="uk-margin">
+                <label class="uk-form-label">Local Destination<span uk-icon="icon: info;" uk-tooltip title="Local accounts that can be used as destination for the swept funds. Make sure you are able to unlock your Nault wallet."></span></label>
+                <div class="uk-form-controls">
+                  <select class="uk-select" [(ngModel)]="myAccountModel" (change)="setDestination(myAccountModel)">
+                    <option *ngFor="let account of accounts" [value]="account.id">{{ account.id }}</option>
```
NiFNi: Shouldn't this use the alias set for the account if there is one?

src/app/components/sweeper/sweeper.component.html
```diff
@@ -0,0 +1,121 @@
...
+                  </div>
+                </div>
+              </div>
+            </div>
+          </div>
+
+          <div class="uk-width-1-1 narrow-div">
+            <div class="uk-form-horizontal">
+              <div class="uk-margin">
+                <label class="uk-form-label">Local Destination<span uk-icon="icon: info;" uk-tooltip title="Local accounts that can be used as destination for the swept funds. Make sure you are able to unlock your Nault wallet."></span></label>
```
NiFNi: Shouldn't the label be assiciated with the input field via the for attribute? Same applies to the other labels.
See here for more information: https://stackoverflow.com/questions/12894169/what-is-the-html-for-attribute-in-label

src/app/services/util.service.ts
```diff
@@ -192,6 +214,17 @@ function stringToUint5(string) {
   return uint5;
 }
 
+function isNumeric(val) {
+  //numerics and last character is not a dot and number of dots is 0 or 1
+  let isnum = /^-?\d*\.?\d*$/.test(val) && val != ''
+  if (isnum && String(val).slice(-1) !== '.') {
```
NiFNi: can be simplified to 
```
return isnum && String(val).slice(-1) !== '.'
```

src/app/services/util.service.ts
```diff
@@ -68,6 +75,15 @@ function hexToUint8(hexValue) {
   return uint8;
 }
 
+// Check if string is hexdecimal
+function isHex(h) {
+  let re = /^[0-9a-fA-F]+$/
+  if (re.test(h)) {
+    return true
```
NiFNi: Can be simplified to 
```
return re.test(h)
```

BitDesert (approved):
Lookin good 👍

NiFNi (commented):
I like the way this works! The UI looks good for me aswell. Some small things came to my mind when looking at the sources. I added those as comments on the files :) Do whatever you want with that :P

src/app/services/util.service.ts
Joohansson: Thanks nifni, I will make some changes before I merge this PR

PR #48, opened 2020-07-08
i18n

### LINKED ISSUE #18: Internationalization support

BitDesert:
Native translations with `@angular/translate` 
3rd party library --> #52 

Desktop builds available as artifacts e.g. https://github.com/BitDesert/Nault/suites/903323004/artifacts/10858459


BitDesert:
@NiFNi Had to pull #35 in a seperate branch to test the Desktop build here, but I think it works flawlessly!
Please test it and give feedback: https://github.com/BitDesert/Nault/releases/tag/v1.5.1
Pinging @Joohansson @aleksre

NiFNi:
@BitDesert I don't have access to my PC until probably tomorrow afternoon (maybe until Saturday) so I can't test the desktop build until then. Your code changes look fine for me though. And the vercel deployment also looks good :+1:

aleksre:
The build works for me in Chrome on desktop (with the preview), but not on my phone and the windows build is stuck on this screen:

image

Opening the preview on my phone, it gets redirected to nault-git-i18n-new.bitdesert.vercel.app**undefined**

Note: My chrome language is set to English, but my Windows and my phone is not. Does it try to auto-detect my langauge? Just a hunch but it might be broken if the users' language does not exist in Nault.

BitDesert:
@aleksre I think I fixed it with e943cc1, there was an issue when the settings were already available.

Joohansson:
@BitDesert The commit you mentioned solved the /en or /de redirection on web but two issues remains:

1. Changing the language and saving the settings result in this (running on a local web server)
image

2. Desktop app on Ubuntu and windows doesn't load. Only a white screen.
image

Seems to be a file error on index.html
image
image

aleksre:
I can't test the desktop build right now but my phone is still getting redirected with the url-suffix `undefined/`. Opening it in incognito works however, so it might have saved a bad settings config to localStorage. Nevertheless, I am unable to clear it because of the instant redirect, We should probably make sure a bad config also redirects to english as a fallback.

NiFNi:
@BitDesert I'm not allowed to push to this branch. Either you allow that or fix it yourself. This should do it:
```
if (settings != null && settings['language'] && Object.keys(routing).includes(settings['language'])) {
```

NiFNi:
Reloading is a bit broken right now aswell (for example try to reload multiple times). I would guess it's by the changes to the base-href but not sure yet.

Joohansson:
Desktop version still white with your change @NiFNi so probably something else. Also same issue as I mentioned before, if you go to the configure app screen and hit refresh you get "entry not found: /en/configure-app"

NiFNi:
@Joohansson The fix was only for the bug @aleksre mentioned ;)

aleksre:
Latest build now works on my phone :) Have not testet the electron build yet.

Can also confirm the refresh bug @Joohansson mentioned. Seems like problem is with redirect.ts line 18: `href.lastIndexOf('/')`

BitDesert:
Closed in favor of #258

PR #55, opened 2020-07-13
Enhanced transaction handling and bug fixes

### LINKED ISSUE #37: The table with incoming tx is messed up on any device not widescreen
**On full screen**

**On any other device (missing account and button to the right)**
image

### LINKED ISSUE #54: Changing rep on a Ledger doesn't work
In certain cases, the rep won't change on a connected Ledger. There is a successful message but nothing changes. This needs to be investigated. Reported for the desktop wallet 1.5.0 (windows).

Joohansson:
Made a few changes for how the user interacts with incoming transactions and solved some bugs.

- Added manual option in "receive priority" setting (incoming will be received manually by the user from the receive-page). Also confirmed to work with the Ledger
- The notification of new incoming is cleared after 10sec instead of forced user interaction. The extra message below the wallet lock should be enough to help the user (Unlock to receive incoming).
- Custom notification messages when manual receive is selected
- Automatically receive incoming (if any) when changing back settings from manual to auto
- Removed bad "issue reference link" in a ledger notification message (old nanovault issue)
- Improved notification warning and success text color to make it actually readable (orange and green)
- Temp fix: receive-table view on small screen (by compacting addresses). Fixes #37
- GUI fix for the "New" label beside receive menu on small devices: Now visible
- Update the receive list automatically when new incoming arrive or incoming received (all accounts or individual)
- Update the account detail list automatically when new incoming arrive or incoming received
- Fixed the bug where representative could not be changed on Ledger device fixes #54

### DIFF
--- a/src/app/components/account-details/account-details.component.ts
+++ b/src/app/components/account-details/account-details.component.ts
@@ -1,5 +1,5 @@
 import {Component, OnDestroy, OnInit} from '@angular/core';
-import {ActivatedRoute, ActivatedRouteSnapshot, ChildActivationEnd, Router} from "@angular/router";
+import {ActivatedRoute, ChildActivationEnd, Router} from "@angular/router";
 import {AddressBookService} from "../../services/address-book.service";
 import {ApiService} from "../../services/api.service";
 import {NotificationService} from "../../services/notification.service";
@@ -47,6 +47,9 @@ export class AccountDetailsComponent implements OnInit, OnDestroy {
   routerSub = null;
   priceSub = null;
 
+  currentPending = [];
+  shouldLoop = false;
+
   constructor(
     private router: ActivatedRoute,
     private route: Router,
@@ -58,6 +61,7 @@ export class AccountDetailsComponent implements OnInit, OnDestroy {
     private wallet: WalletService,
     private util: UtilService,
     public settings: AppSettingsService,
+    
     private nanoBlock: NanoBlockService) { }
 
   async ngOnInit() {
@@ -72,6 +76,32 @@ export class AccountDetailsComponent implements OnInit, OnDestroy {
     });
 
     await this.loadAccountDetails();
+    this.shouldLoop = true;
+    this.updateList();
+  }
+
+  // check for new pending in any account every 2sec and update list if needed
+  async updateList() {
+    let accounts = this.wallet.wallet.accounts;
+    var tempPending = [];
+    let currentAccount = this.router.snapshot.params.account;
+    // save each account in an array to compare next time
+    accounts.forEach(async function(account) {
+      tempPending.push({id:account.id, pending:account.pending});
+      this.currentPending.forEach(async function(entry) {
+        // a certain account pending has changed from last check
+        if (entry.id == account.id && entry.pending.comparedTo(account.pending) != 0 && currentAccount == entry.id) {
+          // reload details
+          await this.loadAccountDetails();
+        }
+      }.bind(this))
+    }.bind(this))
+
+    this.currentPending = tempPending;
+    
+    if (this.shouldLoop) {
+      setTimeout(() => this.updateList(), 5000);
+    }
   }
 
   async loadAccountDetails() {
@@ -135,6 +165,8 @@ export class AccountDetailsComponent implements OnInit, OnDestroy {
     if (this.priceSub) {
       this.priceSub.unsubscribe();
     }
+
+    this.shouldLoop = false;
   }
 
   async getAccountHistory(account, resetPage = true) {
--- a/src/app/components/account-details/account-details.component.ts
+++ b/src/app/components/account-details/account-details.component.ts
@@ -47,8 +47,7 @@ export class AccountDetailsComponent implements OnInit, OnDestroy {
   routerSub = null;
   priceSub = null;
 
-  currentPending = [];
-  shouldLoop = false;
+  statsRefreshEnabled = true;
 
   constructor(
     private router: ActivatedRoute,
@@ -61,7 +60,6 @@ export class AccountDetailsComponent implements OnInit, OnDestroy {
     private wallet: WalletService,
     private util: UtilService,
     public settings: AppSettingsService,
-    
     private nanoBlock: NanoBlockService) { }
 
   async ngOnInit() {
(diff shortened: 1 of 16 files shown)


src/app/services/wallet.service.ts
```diff
@@ -114,23 +114,29 @@ export class WalletService {
         const walletAccount = this.wallet.accounts.find(a => a.id === transaction.block.link_as_account);
         if (walletAccount) {
           // If the wallet is locked, show a notification
-          if (this.wallet.locked) {
-            this.notifications.sendWarning(`New incoming transaction - unlock the wallet to receive it!`, { length: 0, identifier: 'pending-locked' });
+          if (this.wallet.locked && this.appSettings.settings.pendingOption !== 'manual') {
+            this.notifications.sendWarning(`New incoming transaction - Unlock the wallet to receive`, { length: 10000, identifier: 'pending-locked' });
+          }
+          else if (this.appSettings.settings.pendingOption === 'manual') {
+            this.notifications.sendWarning(`New incoming transaction - Set to be received manually`, { length: 5000, identifier: 'pending-locked' });
```
BitDesert: Why is the length here 5 seconds instead of 10?

src/app/services/wallet.service.ts
```diff
@@ -114,23 +114,29 @@ export class WalletService {
...
           await this.processPendingBlocks();
         }
       } else if (transaction.block.type == 'state' && transaction.block.subtype == 'send' && walletAccountIDs.indexOf(transaction.block.link_as_account) !== -1) {
-        if (this.wallet.locked) {
-          this.notifications.sendWarning(`New incoming transaction - unlock the wallet to receive it!`, { length: 0, identifier: 'pending-locked' });
+        if (this.wallet.locked && this.appSettings.settings.pendingOption !== 'manual') {
+          this.notifications.sendWarning(`New incoming transaction - Unlock the wallet to receive`, { length: 10000, identifier: 'pending-locked' });
+        }
+        else if (this.appSettings.settings.pendingOption === 'manual') {
+          this.notifications.sendWarning(`New incoming transaction - Set to be received manually`, { length: 5000, identifier: 'pending-locked' });
```
BitDesert: same here

src/app/components/configure-app/configure-app.component.ts
```diff
@@ -95,6 +95,7 @@ export class ConfigureAppComponent implements OnInit {
   pendingOptions = [
     { name: 'Largest Amount First', value: 'amount' },
     { name: 'Oldest Transaction First', value: 'date' },
+    { name: 'Manual Receive', value: 'manual' },
```
BitDesert: Should we just call it "Manual"? As "Receive" is already in the name of the Setting "Receive Method".

src/app/components/account-details/account-details.component.ts
```diff
@@ -72,6 +76,32 @@ export class AccountDetailsComponent implements OnInit, OnDestroy {
     });
 
     await this.loadAccountDetails();
+    this.shouldLoop = true;
+    this.updateList();
+  }
+
+  // check for new pending in any account every 2sec and update list if needed
+  async updateList() {
```
BitDesert: Is there a certail use case why we need to call `pending` for all account every 5 seconds?
Thought the pending list get's populated by the websocket?

src/app/components/notifications/notifications.component.css
```diff
@@ -15,3 +15,10 @@
   left: 15%;
   z-index: 1050;
 }
+
+.uk-alert-success {
```
BitDesert, line 19: Are those classes not available by default?
https://getuikit.com/docs/alert#style-modifiers

src/app/components/receive/receive.component.ts
```diff
@@ -38,9 +40,46 @@ export class ReceiveComponent implements OnInit {
...
     await this.loadPendingForAll();
   }
 
+  ngOnDestroy() {
+    // stop checking the list when not in this view
+    this.shouldLoop = false;
+  }
+
+  // check for new pending in any account every 2sec and update list if needed
+  async updateList() {
```
BitDesert: Duplicate code, would be better to have that in a seperate file/service.

BitDesert (changes requested):

src/app/services/wallet.service.ts
Joohansson: I thought when using manual receive it's not that important to see the notification on every transaction and maybe 5sec is less annoying but you think it's better to be consistent? I can change to 10

src/app/components/configure-app/configure-app.component.ts
Joohansson: Yeah sure

src/app/components/account-details/account-details.component.ts
Joohansson: Yeah so there are probably 100 ways to do this but I wanted to start with something that works and maybe enhance later. Using this asynchronous method of polling regularly has two advantages as I see it.

1. If you receive 1000 transactions during those 5sec, you only have to call the RPC one time instead of 1000 if it would be a direct connection to the websocket. And the reason for checking pending via RPC is because of 2.
2. By calling RPC and check for pending you get less reliable on the webssocket. It's very important this list is correct and relying on only websocket would make it very easy to miss one transaction if the websocket disconnect just for a few seconds (on either Nault or the backend). The websocket does NOT recover missing transactions. If that would happen now, the list would be completely unaffected and you will always get ALL the missed transactions as soon as the wallet is updated (with an average of 2.5sec). A websocket generally is a really bad solution and doesn't scale very well. There is a reason TIX wallet does not use websocket at all because it's unreliable. So that's the underlying reason for this polling method. You think OK to have for now until we or someone else come up with something better?

src/app/components/notifications/notifications.component.css
Joohansson: Yes but the color suck using the default so I override. I can't read the text so I made it a little darker and more color. Is there a better way?

src/app/components/receive/receive.component.ts
Joohansson: Yeah it's a bit duplicate. I use a faster poll interval here because the list is more essential and what the user use when receiving. The account detail view is more general info. But sure, maybe some kind of service would be better. I'm not sure how to make it but I can try if you are ok with my other comments?

src/app/components/notifications/notifications.component.css
BitDesert: Ah okay, then it's fine.

src/app/components/account-details/account-details.component.ts
BitDesert: My issue is that I want to reduce the calls to the API as much as possible. It's correct that you can miss transactions from the websocket, but then this function should only work as a fallback and should not be called that much to reduce stress on the API. But 5 seconds seems reasonable.

Joohansson:
Everything above should be fixed now

src/app/components/account-details/account-details.component.ts
Joohansson: Yeah, it's been reworked and function much better now. No extra API calls.

src/app/components/receive/receive.component.ts
```diff
@@ -79,38 +56,13 @@ export class ReceiveComponent implements OnInit {
...
-      await this.loadPendingForAll();
-    } else {
-      await this.loadPendingForAccount(account);
-    }
+  async getPending() {
+    // clear the list of pending blocks. Updated again with reloadBalances()
+    this.walletService.clearPendingBlocks()
+    await this.walletService.reloadBalances(true)
+    await this.loadPendingForAll();
   }
```
BitDesert, line 66: Don't we need the functionality to get pending for a specific account?
The old function has both functionalities.

Joohansson: No because it works completely different now. Instead of calling the API to get pending for all accounts or a specific account the data is taken from the already retrieved pending in the wallet service. There is no "burden" to get the full list each time. I filter for specific account directly in the html instead.

BitDesert: Alright 👌

BitDesert (approved):

PR #59, opened 2020-07-14
Backend server stats

### LINKED ISSUE #50: Node information
A [hidden?] page that grabs node information from the RPC could be used for some quick troubleshooting, or general information for those using a custom backend node. Maybe some of them are useful in other pages.

Some useful stats, as usually:
- Block and uncemented count from RPC block_count
- Which network:  main(live) / beta from RPC version
- Telemetry things from RPC telemetry using optionals **address** and **port** to target only the local node information. Unfortunately it would be necesssary to assume (or config option) the local peering port (7075 main net, 54000 beta). Will take a look into potesntially making this simpler for V22.

Joohansson:
A first test of how server info can be displayed. Useful for troubleshooting transactions during network saturation. Fixes #50 

Works for random, pre-configured or custom backend. Stats can be refreshed on the fly but not faster than every 10sec.

image
image

### DIFF
--- a/src/app/components/configure-app/configure-app.component.ts
+++ b/src/app/components/configure-app/configure-app.component.ts
@@ -120,12 +120,12 @@ export class ConfigureAppComponent implements OnInit {
   showServerValues = () => this.selectedServer && this.selectedServer !== 'random';
   showServerConfigs = () => this.selectedServer && this.selectedServer === 'custom';
 
-  nodeBlockCount = 'N/A';
-  nodeUnchecked = 'N/A';
-  nodeCemented = 'N/A';
-  nodeUncemented = 'N/A';
-  nodeVendor = 'N/A';
-  nodeNetwork = 'N/A';
+  nodeBlockCount = null;
+  nodeUnchecked = null;
+  nodeCemented = null;
+  nodeUncemented = null;
+  nodeVendor = null;
+  nodeNetwork = null;
   statsRefreshEnabled = true;
 
   constructor(
--- a/src/app/components/configure-app/configure-app.component.ts
+++ b/src/app/components/configure-app/configure-app.component.ts
@@ -7,7 +7,6 @@ import {PowService} from "../../services/pow.service";
 import {WorkPoolService} from "../../services/work-pool.service";
 import {AddressBookService} from "../../services/address-book.service";
 import {ApiService} from "../../services/api.service";
-import {LedgerService, LedgerStatus} from "../../services/ledger.service";
 import {WebsocketService} from "../../services/websocket.service";
 import {NodeService} from "../../services/node.service";
 import {UtilService} from "../../services/util.service";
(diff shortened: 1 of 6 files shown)


BitDesert (changes requested):
Can we load the Node Stats when changing the server in the dropdown, not only when saving it?
Would be better for comparison and UX, because the other fields (API Server, WS Server etc.) are populated as well when changing the dropdown.

Joohansson:
The server stat updates when switching dropdown now (except for the random server because that is not decided until it's actually being used)

src/app/components/configure-app/configure-app.component.ts
```diff
@@ -137,6 +144,29 @@ export class ConfigureAppComponent implements OnInit {
...
+  }
+
+  async updateNodeStats(refresh=false) {
+    if ((this.serverAPIUpdated != this.appSettings.settings.serverAPI && this.selectedServer === 'random') || (refresh && !this.statsRefreshEnabled)) return
+    this.statsRefreshEnabled = false;
+    try {
+      let blockCount = await this.api.blockCount()
+      this.nodeBlockCount = this.util.string.addCommas(blockCount.count.toString())
+      this.nodeUnchecked = this.util.string.addCommas(blockCount.unchecked.toString())
+      this.nodeCemented = this.util.string.addCommas(blockCount.cemented.toString())
```
BitDesert: Why not use toLocaleString? 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString

This helps us later when we get to translations as well, then we would need to format the numbers according to the locale you chose.

src/app/components/configure-app/configure-app.component.ts
```diff
@@ -113,21 +112,29 @@ export class ConfigureAppComponent implements OnInit {
...
 
   showServerValues = () => this.selectedServer && this.selectedServer !== 'random';
   showServerConfigs = () => this.selectedServer && this.selectedServer === 'custom';
 
+  nodeBlockCount = 'N/A';
+  nodeUnchecked = 'N/A';
+  nodeCemented = 'N/A';
+  nodeUncemented = 'N/A';
+  nodeVendor = 'N/A';
+  nodeNetwork = 'N/A';
```
BitDesert: Why not start with `null` (no value) and update it later with the correct value?
The block counts can be formatted as a number then as well.

BitDesert (changes requested):

src/app/components/configure-app/configure-app.component.ts
Joohansson: Because if the backend server does not support the version call for example I want it to display N/A because that's what it is, not available. If setting it to null, it will either be blank or we have to encode it to N/A in the HTML instead. I figured this method would be easier.

src/app/components/configure-app/configure-app.component.ts
Joohansson: Mostly because the rest of the site is in English. Many countries, Sweden for example, format a large number by using dot instead of comma. Such as 123.456 and that is really confusing when the rest of the site is using dot as a decimal separator. It does make sense when we get translation working so I will change it to use toLocaleString('en-US') for now.

src/app/components/configure-app/configure-app.component.ts
BitDesert: As stated at Discord I would move that to the template anyway, I don't see why this is easier.
You just write `{{ nodeBlockCount || 'N/A' }}` and it's done. Additionally you get strict types (like `Number` for the block counts).

PR #62, opened 2020-07-16
Restyle, mobile fixes and cleanup

zillazillazilla:
This is my first time contributing through GitHub, so if I haven't met common standards that should be then a nudge in the right direction would be appreciated. 

This mostly builds on Joohansson's work [here](https://github.com/BitDesert/Nault/issues/32#issuecomment-652901912)

**General**
- Fixes numerous mobile/tablet usability issues in many components
- Updates the color palette, main font and style
- Corrects some small inconsistencies with buttons and layouts with UIkit

**Bugs**
- Semi-fixes a bug I stumbled across involving the Manage Wallet page where it would infinitely generate errors. A more permanent solution would be to disable the option to Manage a wallet that doesn't exist.


Joohansson:
Nice PR! Looks good to me. Only fix needed as I see right now is the "Import an existing wallet" button on the front page becomes dual line. Would really want that as single line. You can make a new commit on the branch, the PR will change accordingly. A hint for the next time, better if you create a new local branch before making changes. That way you can work on several changes for several PRs. Now you are stuck on the master branch and any changes you make and commit to github will effect this PR.
image

**Edit, another thing.** The address book result has the same background as the background. Needs to have a frame or something or it's very hard to use.

Old:
image

New:
image

zillazillazilla:
Ah - thanks for the heads up and prompt feedback! Looks like I will have to read a bit more into the inner workings of git. Is there anything I can do to address this for now?

All issues should be resolved now.

Joohansson:
Your changes look fine to me. No, you don't need to do anything. It was just a tip =)

aleksre:
This looks great and is very much in line with what I had in mind for a Nault redesign using nano.org as a template 😃  Good job! I haven't found any real issues and have only a few small suggestions
- According to the Nano Guidelines ([brandkit](https://nano.org/brandkit)) they are using Montserrat for headings and 'primary' texts, and Open Sans for body copy. So maybe we should try that as well?
- For a more modern look, I think we should maybe increase the font size on desktop to 18 or even 20px. On mobile it looks fine to me as is. This can easily be done by styling `font-size: 20px` on the root html element inside a media query.
- `uk-button` has a pretty high line-height of 38px. I would decrease this and instead use vertical padding for spacing:
image



As a side note. We should probably use rem values for font-size wherever we need to manually set this with CSS. Right now the nav stands out as not respecting the root font size because it is set with pixels. This is not related to your changes though.

zillazillazilla:
I am just finishing up moving all of the above into .less in order to correctly modify the UIkit theme. The previous commits were a little quick n' dirdy so this should make for a more robust solution in the short term, at least prior to someone tackling a grander redesign! 😉

This change includes all the fixes above and a few more minor alterations for visual bugs.

zillazillazilla:
I appreciate that the scope of these changes are quite extensive so I will conclude this for now. Any further improvements (I want to look at the mobile nav one day) I will leave for a later date should it be merged.

Happy to fix any issues encountered with this pull request, just let me know

BitDesert:
A lot of this PR got into a new PR #70 which we will use from now on to update the design of Nault 💙

PR #64, opened 2020-07-19
Better Account ID readability

aleksre:
This PR utilize the existing `app-nano-account-id` component to prettify nano account id's in the app with colorization of the first and last part of the ID for better readability. It also adds an input option `middle` to the component.

The input option `middle` is used to determine if the middle part of a nano account should be shown. There are 3 options:
- `off`: This is exactly the same as the old behavior of the component, where the middle part is replaced with '...'
- `on`: This will show the middle part between the colorized first and last parts. Essentially exactly the same as simply showing the account directly with `{{ account.id }}` except the beginning and end is colorized for readability.
- `auto`: Applies truncation on the **middle** part of the account id. Basically a slightly more human readable version of `on` when the account doesn't fit on the screen that ensures both the beginning and end is always shown. If the whole account fits then all of it is shown just like `on`. This options wraps the string in a `uk-flex` to determine when truncation should take place so some slight tweaks to the template or style might be needed when this is used inline together with other elements.

I have made `auto` be the default just because it should probably be the most common.

image

image

image

image

### DIFF
--- a/src/app/components/account-details/account-details.component.html
+++ b/src/app/components/account-details/account-details.component.html
@@ -25,11 +25,15 @@
               <div class="uk-width-1-1">
                 <div *ngIf="addressBookEntry">
                     <h3 class="uk-card-title">{{ addressBookEntry }}</h3>
-                    <div class="uk-text-small uk-text-truncate">{{ accountID }}</div>
+                    <div class="uk-text-small uk-text-truncate">
+                      <app-nano-account-id [accountID]="accountID"></app-nano-account-id>
+                    </div>
                 </div>
 
                 <div *ngIf="!addressBookEntry">
-                  <h3 class="uk-card-title uk-text-truncate" style="margin-bottom: 0;">{{ accountID }}</h3>
+                  <h3 class="uk-card-title uk-text-truncate" style="margin-bottom: 0;">
+                    <app-nano-account-id [accountID]="accountID"></app-nano-account-id>
+                  </h3>
                 </div>
               </div>
             </div>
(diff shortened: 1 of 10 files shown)


Joohansson:
I like this and would approve this PR. I would like to see it also in the send confirmation screen (with middle=auto) where both FROM and TO accounts are shown.

image

aleksre:
Good point. I'll add it there as well.

aleksre:
Added to send confirmation. Also made it possible to add classes to the flex wrapper from outside the component.

BitDesert:
I think another lighter color for the `:hover` event is still missing.
And can we include 5 chars after the `nano_` prefix? Then it's the same number of chars at the beginning and the end.
Found another spot where we need to change it: At `/manage-representatives`

aleksre:
@BitDesert When you say hover event do you mean on the component itself or when it's being used as a link?

I'll fix the other two issues.

Joohansson:
I saw on the send confirmation for the new account feature, there was a space between the mid section ... and the final end section. Seems to be dependent on the screen width so like 50% chance the space will be too large for looking good. Is that possible to solve? To close the gap I mean.

BitDesert:
@aleksre Yep, the color when used as a link: 
grafik
grafik

aleksre:
@BitDesert Done. Had to update the link hover styling itself. I use a combination of color change and brightness filter so it should work for any color.

@Joohansson Oh, ok I see what you mean. That happens because there barely isn't enough room for another character so in some cases it might appear as a small space before the last part. I tried to fix it, but since the ellipsis is a result of its container shrinking in size, I don't think there is a way to do it this with CSS. It might be possible with JavaScript, but it will be hacky and I would rather avoid that.

PR #67, opened 2020-07-21
My Nano Ninja Integration

### LINKED ISSUE #11: Manage known reps only show NF reps
Maybe show the suggested list of reps instead?

### LINKED ISSUE #14: Improved workflow for switching out a bad rep
Now you get a warning and has to click on that, then select the right account and a rep from a list. It's many steps for the normal user. I'm thinking if it can be done more semi-automatic with the wallet taking the action and user just approve. I think that will help with weight distribution.

For example a dialog popping up:

> Your account’s representative is not optimal due to high weight. Press OK if you allow Nault to change it for you to rep XYZ. Press MANUAL if you want to select the rep yourself. Or CANCEL if you want to keep a bad rep.

BitDesert:
This PR integrates an API to My Nano Ninja for pulling rep information into Nault.
It merges the local known reps with the reps known by Ninja to display the correct information.


aleksre:
Looks very nice. Should we also show some additional metrics in the dropdown that will make picking a good rep easier such as score, weight and uptime? Would also really like to see even more metrics such as location and node version here. I personally want to pick a rep that is fairly close to me.

We should probably also consider removing or hiding the "View recommended representatives" list. Since it tries to solve the same use-case but doesn't support search.

BitDesert:
I see the dropdown more like a autocomplete and not as a fully functioning list. That's why I left the "View recommended reps" list as is.
If some information is missing we can add it there as well, but it shouldn't be too cluttered. If you want all information you should visit it's profile at mynano.ninja

Joohansson (approved):
I think it looks great!
Not required but the text may be improved. For example at the bottom here, maybe make it more clear that a rep has been found but not exclusive.

**Example:**
A new representative has been suggested for you by the system: REPNODE
Click "CHANGE" to approve or "MANUAL" to select another.

image

After clicking change the user is required to click another button to actually make the change. This may have to be more clear or do that automatically. But I guess the account(s) need to be selected first.

I would change the title below to "Recommended"
image

BitDesert:
The list is not really a "recommended" list but more like a autocomplete for the input.
Added the text for the "manual" button.

PR #70, opened 2020-07-23
Redesign

BitDesert:
Redesign of @keerifox and @zillazillazilla


aleksre:
Looks fantastic! 😁 

My thoughts:

The idea to visualize each account as a credit card works really well and helps visualize the concepts to the user. Especially the wallet-account relationship. I think we should just be careful with using the card on the "Total balance" for the entire wallet. We should make sure we are giving the user an accurate and consistent mental model:

> - A Nano wallet holds one or more accounts, each represented by a virtual credit card
> - This is consistent with what they already know from their physical wallet which also hold several credit cards
> - It is also consistent with their fiat bank app where after logging in, the money is stored in multiple accounts (each of which may also have a credit card associated with it)

We should not give the impression that the *wallet itself* is a credit card or can hold/send/receive nano

So I suggest we always show an account in the nav. The total wallet balance can be displayed above it (if it is even needed? It is also shown in the dropdown)

Something like this:
image

The total balance could perhaps also be hidden when you only have a single account.

The account dropdown looks great btw 🙂 

When the wallet is locked, maybe we should grey out the card or something? I feel like the app is missing a visual cue when this happens.

Future improvment suggestion: Give the option to set a color for each card so the accounts are more easily distinguishable

Sidenote: The new tooltip (representative health) looks great. Can we make this the new default?

keerifox:
Thanks!

> We should not give the impression that the wallet itself is a credit card or can hold/send/receive nano

wallet can hold/send/receive nano so i don't think there's any harm for showing it on the card. there isn't any additional space to fit yet another balance and while it would be possible to make it look like a stack of cards when displaying total balance, i'm not sure if that's really necessary

aleksre:
Only accounts can hold/send/receive nano. Wallets contain one or more accounts. This is precisely the confusion I want to avoid. It is not at all obvious to newcomers.

What, in your opinion, does the card represent then? Is it simply a stylized container with no meaning?

I'm not sure I understand what you mean by additional space. We don't need any more space to solve this. Just default to showing Account #0.

keerifox:
i still don't think there is any harm for having the same look for wallets and individual accounts but if you really wanted to, you could have it appear like a wallet when displaying total balance

image

```css
.nav-nano-card.is-wallet {
    background: #506279;

    > .logo {
        filter: saturate(0.5) brightness(0.70);
    }

    &::after {
        content: '';
        position: absolute;
        width: calc(100% - 10px);
        height: calc(100% - 10px);
        display: block;
        border: 2px dotted #1F1F26;
        margin: 3px;
        border-radius: 10px;
        pointer-events: none;
    }
}
```

i'm not sure how i feel about defaulting to account #0 considering every NanoVault/Nault user is already used to "total balance" being the default, and there will be some confusion when their funds will not appear in the balance due to them looking at a specific account. if they switch it themselves at least once, they are more likely to remember to switch to "total balance" or the account they want to see receives on.

Joohansson:
I think the rep view should be compacted by default because if you have many accounts you will have to scroll just to find the actual wallet functions. Is it THAT important to show them all the time? Sure this is an extreme case, but just a thought.
image

keerifox:
my implementation had only shown one rep (which was meant to display rep of the currently selected account)

for the "total balance" view i suggested that it shows only one rep of your account with the most confirmed balance

BitDesert:
IMO if you have multiple accounts and choose a different representative for _every_ account this view is fine.
Most people will only see 1 rep anyway, for those who have multiple I assume they delegated to not more than 3 or so. And even that's a lot.

aleksre:
> i'm not sure how i feel about defaulting to account #0 considering every NanoVault/Nault user is already used to "total balance" being the default, and there will be some confusion when their funds will not appear in the balance due to them looking at a specific account. if they switch it themselves at least once, they are more likely to remember to switch to "total balance" or the account they want to see receives on.

This isn't really an issue. We are redesigning the app, so it is totally fine and even expected that the location of elements will be changed around as well. It is much more important that we make sure these complex concepts are presented in a consistent and easy to understand manner for people not familiar with crypto. Showing the wallet instead of the card for Total Balance is a step in the right direction, but could still be misunderstood I think. Having Total Balance shown only inside the dropdown is more than enough IMO.

@Joohansson I agree. It makes sense to only show the representative for the currently selected account. Once again the issue is what to do when "Total Balance" is selected.

Joohansson:
Could also show reps as default IF they are less than let's say 3. If more than 3, display a button to expand the rest. That is part of the design to solve that logic IMO.

keerifox:
i asked a few community members and they seemed to be in favor of adding distinct look for wallet view (total balance). i've improved the logo contrast a bit and adjusted border to look more appealing

image


```css
.nav-nano-card.is-wallet {
    background: #506279;

    > .logo {
        filter: saturate(0.5) brightness(0.75);
    }

    &::after {
        content: '';
        position: absolute;
        width: calc(100% - 10px);
        height: calc(100% - 10px);
        display: block;
        border: 1px dashed #1F1F26;
        margin: 4px;
        border-radius: 10px;
        pointer-events: none;
    }
}
```

aleksre:
All right, let's go with that then 👍 

Btw, is the intention that selecting an account in the nav will "select" this account globally for the app? We currently have individual account dropdowns on several pages like Send and Receive and these could be simplified if we use the selected account. This is how Natrium does it and I think it works well, but not sure if this is the intention or not?

keerifox:
all app dropdowns could default to the currently selected account, absolutely

BitDesert:
What's missing here? Would like to merge it so we can go on with the translations.

BitDesert:
@keerifox The rep list dropdown text is white on white:
grafik

keerifox:
> What's missing here? Would like to merge it so we can go on with the translations.

a state containing basic information about the currently selected account (with balance/fiat/rep/address), which is either a nano_ account or the entire wallet. account selection dropdown should have the active acc at the top and the rest underneath it (most code for it is there). clicking an item should change to that account. wallet and accounts cards look differently (code also there just need to pass a variable). when an account is active, shown rep should be the rep of that account; clicking the name should offer to change rep of currently selected nano_ address (2 todos in change rep widget). i think that's all

non-critical: instead of "Account X" it should say the label in the address book, on both card and the account selector dropdown

would be nice if a few people could test it after that's added

keerifox:
> 
> 
> @keerifox The rep list dropdown text is white on white:
> grafik

i think it's been like that in zill's fork, just add this somewhere in src/less/nault-theme.less

```css
.uk-card-body > ul > li > a {
    color: #676686;
}
```

BitDesert:
Looks fine for me for merge 👍

Joohansson (approved):
Looks fine for me as well. I only noticed two things but I can live without them.

1. There is no "home button" to get to the start screen once you left it. In the old design, you could click the logo
2. When changing account in the menu, maybe would be more convenient to also change the default account chosen when sending? Now it defaults to the first account. The account chosen is also not saved when refreshing. This has nothing to do with the design but I think it should be built later.

PR #75, opened 2020-07-26
redesign: improved handling of representatives and manual receives, wallet style for total balance view

keerifox:


BitDesert:
@keerifox there seems to be an issue:
```
ERROR in src/app/components/change-rep-widget/change-rep-widget.component.ts:11:14 - error TS2420: Class 'ChangeRepWidgetComponent' incorrectly implements interface 'AfterViewInit'.
```
Can you solve that? Then we can merge it, Thanks!

Issue #72, opened 2020-07-25
Error: Cannot find module './desktop-app/src/lib/ledger'


iljah:
Installed from source as described in documentation:
```
...
npm install
npm run wallet:build-desktop
npm run desktop:local
```
on fedora 32 and tried to run but got:
```
$ desktop-app/build/Nault-1.5.2.AppImage   
Starting ledger@!
A JavaScript error occurred in the main process
Uncaught Exception:
Error: Cannot find module './desktop-app/src/lib/ledger'
Require stack:
- /tmp/.mount_Nault-vFwEN9/resources/app.asar/main.js
- 
    at Module._resolveFilename (internal/modules/cjs/loader.js:797:17)
    at Function.o._resolveFilename (electron/js2c/browser_init.js:281:679)
    at Module._load (internal/modules/cjs/loader.js:690:27)
    at Function.Module._load (electron/js2c/asar.js:769:28)
    at Module.require (internal/modules/cjs/loader.js:852:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.<anonymous> (/tmp/.mount_Nault-vFwEN9/resources/app.asar/main.js:12:16)
    at Module._compile (internal/modules/cjs/loader.js:967:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1004:10)
    at Module.load (internal/modules/cjs/loader.js:815:32)
```
and nothing else seems to happen. Is something wrong or should I just wait longer?

BitDesert:
If your build didn't get any errors this seems to be weird...
Is the prebuilt AppImage working for you? https://github.com/BitDesert/Nault/releases/download/v1.6.1/Nault-1.6.1.AppImage

iljah:
How do I convert checksum in yml to hex?

BitDesert:
I'm not familiar how this works in Fedora. You probably have to find that out yourself or ditch the checksum for testing purposes.

iljah:
This isn't related to fedora because release page writes `shasum -a 512 <filename>` which gives: `354ba4e79789a8cd082994239c7eb61620246110c0a3f6a9a4cbe6c2ec4a24bf78420012ef29605d01b4bfbc305fc124f5183d628fd71b00dbcea78e9bc7f398  Nault-1.6.1.AppImage`

[...]

PR #77, opened 2020-07-27
Electron cleanup

BitDesert:
There are some probably leftover files from a test by cronoh which should get deleted.
Also the docs for the desktop build and the npm scripts got cleaned up.

Related to #72


BitDesert:
@Joohansson Can you test the electron build by compiling it yourself and from the artifacts?
I tested it myself and the Ledger still worked but I want to be sure. Thanks!

Joohansson (approved):
All seems fine for both self-compile and running the artifact. Tested on windows and doing all Ledger functions.

PR #102, opened 2020-07-29
Reusable QR scanner modal

Joohansson:
QR scanner modal to be used anywhere in the wallet. Using unique ID for example text inputs to match the correct value scanned.

- QR modal that allows any dynamic component to be loaded with a special service to minimize duplicate code
- Added camera icon to most input fields: account, rep, seed, key, mnemonic
- Added ninja service to the default rep in the app settings
- Added hard route for private key when on format "nanokey:" (default QR scanner)

Default QR scanner unchanged.
image


package.json
```diff
@@ -34,13 +34,16 @@
     "@angular/compiler": "^9.1.11",
     "@angular/core": "^9.1.11",
     "@angular/forms": "^9.1.11",
+    "@angular/localize": "^9.1.12",
```
BitDesert, line 37: Why do we include localization here?

package.json
```diff
@@ -34,13 +34,16 @@
...
     "@angular/platform-browser": "^9.1.11",
     "@angular/platform-browser-dynamic": "^9.1.11",
     "@angular/router": "^9.1.11",
     "@ledgerhq/hw-transport-node-hid": "^5.19.0",
     "@ledgerhq/hw-transport-u2f": "^5.19.0",
+    "@ng-bootstrap/ng-bootstrap": "^7.0.0",
```
BitDesert, line 43: Any special cause why we need Bootstrap when we're using UIkit? 
How much larger is the rendered JS/CSS?

Joohansson: Yeah so I was following this tutorial: https://medium.com/better-programming/dynamically-loaded-bootstrap-4-modal-component-powered-by-angular-6621f5fd3173

Specifically the "this.modalService.open(QrModalComponent)" is using NgbModal from the bootstrap lib. I actually searched for hours to find a similar way with UIKit but it doesn't seem possible to load another component into a modal with it, or no one has documented it. So that's why. I needed a reusable component. Or the QR scanner logic would have to be duplicated into every view that needed it.

The final JS/CSS increased 100kB. From 3.8 to 3.9 MB. Compared to the master branch.

Joohansson: Actually, I could remove ajv. Don't know why the guide was needing that.

package.json
Joohansson: After I installed @ng-bootstrap/ng-bootstrap": "^7.0.0" it was needed for Angular 9 or Nault would not load.

PR #116, opened 2020-08-01
Upgrade to Angular 10

### LINKED ISSUE #113: Angular 10 Update
It would be good when we would use the newest stable Angular version (if compatible).

aleksre:
Upgraded Nault to Angular version 10.

Everything seems to be working. Please test it and verify.

There are not that many breaking changes in Angular 10, but here are some things to note:
- `<input type="number">` no longer responds to 'change' events. After seaching around the app it looks like we are not using this at all.
- From what I understand EntryComponents are no longer needed since Ivy. I removed it and checked that the qr modal still works, so unless I'm missing something this should be fine.
- Angular 10 will print a lot of warnings in the console about the use of non-ECMAscript module dependencies (which can't be optimized). These still works just fine, but it's probably a good idea to use tree-shakable modules as much as possible going forward, and transitioning our existing ones if possible. Thus I decided not to disable these warnings. In some cases the way we import the modules might cause them to go away, but I considered this out of scope for the upgrade.

Solves #113


BitDesert (approved):
Tested basic functionality, works properly!

Joohansson (approved):
Installed npm and compiled from scratch. Tested core functions and ledger both web and desktop. Works fine as far as I can tell.

PR #121, opened 2020-08-02
Ledger Auto-Import

### LINKED ISSUE #120: Auto-Scan for funds when using Ledger
Currently Nault only adds the first account when you use a Ledger device.
It would be better, when it would scan the first 10 accounts for funds like when using a seed.

You don't have to accept that action, so no UX issues there.

The code should be shared between seed/ledger.

BitDesert:
Fixes #120


Joohansson (approved):
Works fine, found all my 6 accounts. However, there is a 5 sec delay before it adds the accounts, why is that?

BitDesert:
The ledger communication seems a bit slow, that's why it takes some time.

PR #123, opened 2020-08-02
Fix unknown reps not reporting 0% weekly uptime

### LINKED ISSUE #115: Misleading rep status for offline reps (not immediately known by MyNanoNinja?)
image

choosing your own address as a representative tells you it's good until you refresh the page (which might be too late for people that decide it is a good idea but won't visit the wallet again in a while)

how do we proceed with this?

1a. there could be a neutral "Unknown Representative" status
1b. it could default to "Bad Representative" as we don't know anything about it
1c. re-fetch information in a couple seconds? [there should already be a style](https://i.imgur.com/dAdmgR1.png) for "Checking Representative..." somewhere in my commit history

2a. replace "It is highly advised to change..."  with "We were unable to determine the status of this representative."
2b. add an error "This representative was <b>never seen online</b>" or something along the lines

keerifox:
This fixes #115, however i'd like to make sure 0% weekly uptime is only shown if the ninja service returns a 404 error and not simply fails the api call for any other reason. Otherwise we could risk reporting 0% weekly uptime on good reps.

We might still need an "Unknown Status", e.g. if the api call returns any other error code, as in that case we can't rule out "good" nor "bad".

### DIFF
--- a/src/app/components/change-rep-widget/change-rep-widget.component.html
+++ b/src/app/components/change-rep-widget/change-rep-widget.component.html
@@ -13,10 +13,14 @@
         : ''
       }}</h2>
     <div class="representative-name-row">
-      <a (click)="showRepSelectionForSpecificRep(rep)" class="name">{{ rep.label || 'Unknown' }}</a>
-      <div class="weight-total">{{ rep.percent.toFixed(2) }}%</div>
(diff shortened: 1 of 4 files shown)


keerifox:
This should be properly solved now

Joohansson (approved):
Don't know how to test it but code looks good to me.

Issue #105, opened 2020-07-29
Linting


BitDesert:
It would be great to have a common linter which would run for every commit/PR.
That way the codestyle is defined in some config files and everyone could write code that has the same style as from everybody else.

- [x] Define which linter to use
- [x] Initial linting + fixes
- [x] Implement action for every push + PR

sebrock:
Angular´s built-in **ng lint** uses **tslint**
https://github.com/Nault/Nault/blob/d7d9a46dbdbcdd9212b4e1338182ba959a33ba3f/package.json#L87
**tslint has been deprecated**, replacement:[ typescript-eslint](https://eslint.org/) [github](https://github.com/typescript-eslint/typescript-eslint#typescript-eslint)
And there is [angular-eslint](https://github.com/angular-eslint/angular-eslint)

sebrock:
For less, there is [lesslint](https://github.com/lesshint/lesshint)

Joohansson:
Ok working on the initial linting + fix part

Issue #112, opened 2020-07-30
Bug: Account does not update after importing a new wallet


aleksre:
Steps to reproduce:

1. Import a wallet
2. Select an account in the nav
3. Import another wallet (I imported from a ledger but don't think it matters)

**Expected:** Total balance should be selected in the nav (or an account from my newly imported wallet)
**Actual:** The old account is still selected and shown

Both the account and the rep shown here is from the first wallet, after importing from the ledger.
![image](https://user-images.githubusercontent.com/1097444/88982485-ae6ea380-d2c8-11ea-897d-ff8197384f83.png)

The dropdown shows both my Primary Account (old wallet) and Account#0 (new wallet).
![image](https://user-images.githubusercontent.com/1097444/88982617-14f3c180-d2c9-11ea-8e67-383ff4fda62d.png)

Joohansson:
Solved with commit 0d6e609 and 4e88ca3 which is part of #125

PR #125, opened 2020-08-02
Code refactor and linting

Joohansson:
Ran "ng lint --fix" and then started doing some manual fixing, but the linter should be set up correctly before continuing with the rest 400 errors. Some rules can maybe be ignored. I'll pass this on for now.

Linked: #105


aleksre:
Looks good 👍 

TSLint is deprecated though. Should we transition to ESLint?

BitDesert:
Definitely @aleksre ! 
We might use something like https://github.com/angular-eslint/angular-eslint or do you have any suggestions?

According to https://github.com/angular/angular-cli/issues/13732#issuecomment-669145618 ESlint will come be implemented to Angular 11, so maybe we should wait too.

aleksre:
Perfect! My only suggestion would be that we enable auto-fix on save in VS Code by adding a `.vscode/settings.json` file with:
```
"editor.codeActionsOnSave": {
    "source.fixAll.eslint": true
}
```
Whether we add angular-eslint or wait for Angular 11 I don't really have a strong opinion. Both seems like fine options. I'd imagine we anyway need some time to tweak our rules though.

BitDesert:
I think the rules are now okay to work with and I did the first automatic fix run.
Now I'm starting to fix the rest manually.

Is that auto-fix working for tslint as well?

aleksre:
Yeah, I meant the eslint rules (since they are completely different from tslint)

Actually, if we instead add this:
```
"editor.codeActionsOnSave": {
    "source.fixAll": true
}
```
It should, at least in theory, apply fixes for all linters, so it should work for both TSLint and ESLint 😁 
Dont know if it works but you can give it a try. Make sure you have the [TSLint plugin](https://marketplace.visualstudio.com/items?itemName=ms-vscode.vscode-typescript-tslint-plugin) installed.

BitDesert:
@aleksre can you check the following errors? Some of them got introduced with daa6746740efc99123050d11615f9e48ebaff17c but I don't know exactly how they work. Thanks!

src/app/components/helpers/nano-account-id/nano-account-id.component.ts:7:3
ERROR: 7:3    no-host-metadata-property
Use @HostBinding or
@HostListener rather than the `host` metadata property (https://angular.io/styleguide#style-06-03)
  
ERROR: 13:3   no-input-rename
In the class "NanoAccountIdComponent", the directive input property "accountID" should not be renamed.
However, you should use an alias when the directive name is also an input property, and the directive name
doesn't describe the property. In this last case, you can disable this rule with `tslint:disable-next-line:no-input-rename`.
  
ERROR: 14:3   no-input-rename
In the class "NanoAccountIdComponent", the directive input property "middle" should not be renamed.
However, you should use an alias when the directive name is also an input property, and the directive name
doesn't describe the property. In this last case, you can disable this rule with `tslint:disable-next-line:no-input-rename`.

aleksre:
@BitDesert Sure. Just replace the component with this:
```
import {Component, Input, OnChanges, HostBinding} from '@angular/core';

@Component({
  selector: 'app-nano-account-id',
  templateUrl: './nano-account-id.component.html',
  styleUrls: ['./nano-account-id.component.css'],
})
export class NanoAccountIdComponent implements OnChanges {

  @HostBinding('class.uk-flex') isFlex: boolean;
  @Input() accountID: string;
  @Input() middle: 'on'|'off'|'auto' = 'auto';

  firstCharacters = '';
  middleCharacters = '';
  lastCharacters = '';

  constructor() { }

  ngOnChanges() {
    this.isFlex = this.middle === 'auto';
    const accountID = this.accountID;
    const openingChars = 10;
    const closingChars = 5;
    this.firstCharacters = accountID.split('').slice(0, openingChars).join('');
    this.lastCharacters = accountID.split('').slice(-closingChars).join('');
    if (this.middle !== 'off') {
      this.middleCharacters = accountID.split('').slice(openingChars, -closingChars).join('')
    }
  }

}
```

BitDesert:
Ready for review, all basic frunctionalities have to be tested as many comparisions are now checked strictly with `===` instead of just `==`.

Tested working:
- Existing wallet
- Send/Receive (seed)
- New wallet generation (seed)

Joohansson:
**Tested desktop app:**

- Sweeper tool
- New wallet, Import seed
- Send/receive/change rep using Nault
- Send/receive/change rep using Ledger
- Send/receive/change using remote signing
- Scan QR
- Manual receive
- Change server

**Changes:**

- Display QR in the receive screen if a default account has been selected in the sidebar
- Include selected default account in local store. Load it back when reloading Nault
- Send, receive and sweeper has account index in their dropdowns. Now also in the representative screen
- Some more account-copy-button placement fixes when screen is narrow
- Solved #112 (part of commit 0d6e609 and 4e88ca3)

PR #127, opened 2020-08-04
Card headings links

keerifox:
Total Balance links to /accounts
Account Name links to /account/`id`
Incoming Balance links to /receive


Joohansson (approved):
Works fine!

PR #128, opened 2020-08-04
Badges

BitDesert:
Just some small badges on top of the readme with some quick links and information which is not directly available otherwise.


Joohansson (approved):
Looks good. The 2.3k downloads, is that with NanoVault included?

BitDesert:
I don't think it's with NanoVault included, as this repo is not flagged as being a fork.
But this number could be a bit high because the desktop app checks for updates on every start. NPM for example counts for every HTTP 200 request.

PR #130, opened 2020-08-05
fix(GUI): all copy/edit buttons visible without hover

### LINKED ISSUE #111: Easier way to copy address on mobile
Currently the "Copy to Clipboard" button only appears on hover, could it be always visible in mobile view? (or desktop too)

### LINKED ISSUE #124: Adding labels to own addresses from Accounts page
At the moment it's very easy to miss out on the information that you can label your own addresses, as you have to visit the address book page to see the hint.

Accounts should have a pencil icon next to them with hover text "Change Label" for quick access to creation of a new address book contact, with the address prefilled. One small issue is there is no functionality for changing labels, only removing/creating new.

Joohansson:
Copy address buttons was only visible on hover which was not only annoying but also impossible to use on mobile
- Removed all "uk-hidden-hover" styles

Fixes #124 and fixes #111 

Question: Is this ok UI-wise or too many visible buttons in tables, etc.? It's ok IMO.

### DIFF
--- a/src/app/components/account-details/account-details.component.html
+++ b/src/app/components/account-details/account-details.component.html
@@ -23,10 +23,10 @@
             </div>
             <div *ngIf="!showEditAddressBook" uk-grid>
               <div class="uk-width-1-1">
-                <div *ngIf="addressBookEntry">
+                <div *ngIf="addressBookEntry || walletAccount">
                   <div class="uk-width-1-1">
                     <div uk-grid>
-                    <h3 class="uk-width-auto uk-card-title uk-text-truncate" style="max-width: calc(100% - 35px);">{{ addressBookEntry }}</h3>
+                    <h3 class="uk-width-auto uk-card-title uk-text-truncate" style="max-width: calc(100% - 35px);">{{ addressBookEntry ? addressBookEntry : ('Account #' + walletAccount.index)}}</h3>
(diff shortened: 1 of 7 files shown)


keerifox:
i would move the buttons as close as possible to the text they are related to

e.g. it's good on the receive page
image

but on the account page both buttons are far from account label/address they apply to
image


it's a good start though, much better than hidden buttons!

keerifox (approved):

BitDesert (changes requested):
I would move the buttons always right after the text which is copied (e.g. address). Otherwise you don't know to which text the copy button belongs to.

Joohansson:
Did some tightening of the positions. Now the edit button also switches place if there is an account label.

image
-------------
image
-------------
image
-------------
image

I left the space in the account overview because the buttons would not align otherwise and it would look bad I think.
image

keerifox:
image

could it say "Account #X" here instead of "Awesome account", when no label is set?
would remove ambiguity from the pencil icon, and make it easier to connect accounts in the sidebar with accounts in the main view

Joohansson:
> could it say "Account #X" here when no label is set?
> would remove ambiguity from the pencil icon, and make it easier to connect accounts in the sidebar with accounts in the main view

Done!
image

PR #76, opened 2020-07-27
Multicurrency Support

BitDesert:
Enables support for custom themes and functionalities for every Nano fork (e.g. Banano).

PR #131, opened 2020-08-05
Warn for 2%

BitDesert:
IMO a "warning" for a rep should occur at >2%.
If a rep has ~1.X % this should not trigger a warning.

At the current state this would be better than the current check, as the decentralization get's better these limits can get modified again.

### DIFF
--- a/src/app/components/change-rep-widget/change-rep-widget.component.html
+++ b/src/app/components/change-rep-widget/change-rep-widget.component.html
@@ -49,7 +49,7 @@
         <span uk-icon="icon: warning;"></span>This representative has a <b>very high voting weight</b> (over 3%).
       </p>
       <p class="uk-text-warning" *ngIf="rep.status.highWeight">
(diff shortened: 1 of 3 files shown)


keerifox:
sounds ok to me, just remember to change 1% to 2% in these files:
change-rep-widget.component.html
representatives.component.html

Joohansson (changes requested):
As keeri said, please change the text as well so it's not forgotten:

- https://github.com/Nault/Nault/blob/05b9272f46e489a76502e0a29ea8c2b4e8894036/src/app/components/representatives/representatives.component.html#L37
- https://github.com/Nault/Nault/blob/05b9272f46e489a76502e0a29ea8c2b4e8894036/src/app/components/change-rep-widget/change-rep-widget.component.html#L56

aleksre:
Would maybe be a good idea to make these global constants and reference them in both code and texts so they can be easily adjusted as decentralization improves.

BitDesert:
I will port that to the environtment with #76 and just fix it in the code for now.

Joohansson (approved):

PR #133, opened 2020-08-06
docs(security): Links to the security docs

Joohansson:
-Added link on the remote signing page and home screen
-Hidden info page by default on the home screen

### DIFF
--- a/src/app/welcome/welcome.component.html
+++ b/src/app/welcome/welcome.component.html
@@ -38,7 +38,7 @@
           <div class="uk-card-body">
 
             <ul uk-accordion="">
-              <li>
+              <li class="uk-open">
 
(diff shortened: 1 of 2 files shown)


src/app/welcome/welcome.component.html
```diff
@@ -38,15 +38,15 @@
           <div class="uk-card-body">
 
             <ul uk-accordion="">
-              <li class="uk-open">
```
BitDesert: We should keep it open until we redesign the homepage.

BitDesert (changes requested):

src/app/welcome/welcome.component.html
Joohansson: Restored

PR #143, opened 2020-08-10
Save last price

### LINKED ISSUE #141: Cache last price
It would be better, when the last price would be stored in the browsers local storage.
That way the price does not jump that much around when the app is loaded and is failsafe when the external service is not available.

BitDesert:
Saves last price in local storage for caching.

fixes #141


Joohansson (approved):
Looks good

PR #158, opened 2020-08-16
feat(qr-reader): Ability to scan an exported wallet/address book

Joohansson:
A wallet can be exported as a QR. That can be read by mobile for example to link to nault.cc and import the wallet. By allowing full web paths to be read, also the nault QR-reader can be used for the same QR.

Bonus: Works with any http or https URL like https://google.com
Could use Nault as a generic QR scanner on any device.

Edit:
Bad idea to allow generic paths and reloading. Now using only internal routes.

### DIFF
--- a/src/app/components/qr-scan/qr-scan.component.ts
+++ b/src/app/components/qr-scan/qr-scan.component.ts
@@ -75,10 +75,10 @@ export class QrScanComponent implements OnInit {
       if (url.protocol === 'https:') {
         if (url.pathname === '/import-wallet' && url.hash.slice(1).length) {
           // wallet import
-          this.router.navigate(['import-wallet'], { fragment: url.hash.slice(1)});
+          this.router.navigate(['import-wallet'], { queryParams: {hostname: url.hostname}, fragment: url.hash.slice(1)});
         } else if (url.pathname === '/import-address-book' && url.hash.slice(1).length) {
           // address book import
-          this.router.navigate(['import-address-book'], { fragment: url.hash.slice(1)});
+          this.router.navigate(['import-address-book'], { queryParams: {hostname: url.hostname}, fragment: url.hash.slice(1)});
         }
       } else if (url.protocol === 'nano:' && this.util.account.isValidAccount(url.pathname)) {
         // Got address, routing to send...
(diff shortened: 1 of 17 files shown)


src/app/components/qr-scan/qr-scan.component.ts
```diff
@@ -71,6 +71,12 @@ export class QrScanComponent implements OnInit {
       // This is a valid Nano scheme URI
       const url = new URL(resultString);
 
+      // check if QR contains a full web URL. For example a full wallet import
+      if ((url.protocol === 'http:' || url.protocol === 'https:')) {
+        window.location.href = resultString;
```
BitDesert: This leads to some redirection issues.

BitDesert (changes requested):

src/app/components/qr-scan/qr-scan.component.ts
Joohansson: Should be better now

src/app/components/qr-scan/qr-scan.component.ts
```diff
@@ -71,7 +71,16 @@ export class QrScanComponent implements OnInit {
...
-      if (url.protocol === 'nano:' && this.util.account.isValidAccount(url.pathname)) {
+      // check if QR contains a full URL path
+      if (url.protocol === 'https:') {
+        if (url.pathname === '/import-wallet' && url.hash.slice(1).length) {
+          // wallet import
+          this.router.navigate(['import-wallet'], { fragment: url.hash.slice(1)});
+        } else if (url.pathname === '/import-address-book' && url.hash.slice(1).length) {
+          // address book import
+          this.router.navigate(['import-address-book'], { fragment: url.hash.slice(1)});
+        }
```
BitDesert: Here is missing the check for the current domain.
The domain in the QR code has to match with the domain you're on.
Otherwise you could scan `https://thisisnotnault.cc/import-wallet#abc` and it would work which is probably not intended.

BitDesert (changes requested):

src/app/components/qr-scan/qr-scan.component.ts
BitDesert: Okay it's fine as we only take a part of the domain, tampering with Nault isn't possible with this approach.

BitDesert (approved):

PR #161, opened 2020-08-16
Inform user when incoming tx list is loading

keerifox:
Adds this loading label (when fetching account frontiers or upon pressing "Find Incoming" button)

image

Also ensures the list is cleared when you press the "Find Incoming" button


Joohansson (approved):
Works fine, thanks!

PR #163, opened 2020-08-17
fix(sidebar): No more flickering when refreshing incoming

### LINKED ISSUE #160: Incoming balance appears in sidebar then disappears
How to reproduce:
1. In the sidebar make active an account that has incoming tx, all of which are below minimum receive amount (1 tx is enough)
2. Press "Find Incoming" button on the receive page or simply refresh the page

this is likely to walletAccount.pending getting a new value
https://github.com/Nault/Nault/blob/da4bc6311b1cacd5935faff81ec9f6c79f83eef7/src/app/services/wallet.service.ts#L717

before minimums are processed
https://github.com/Nault/Nault/blob/da4bc6311b1cacd5935faff81ec9f6c79f83eef7/src/app/services/wallet.service.ts#L760

Joohansson:
If having incoming below min threshold
Fixes #160

### DIFF
--- a/src/app/services/wallet.service.ts
+++ b/src/app/services/wallet.service.ts
@@ -24,6 +24,7 @@ export interface WalletAccount {
   balance: BigNumber;
   pending: BigNumber;
   pendingBelowThreshold: BigNumber[];
+  pendingOriginal: BigNumber;
   balanceRaw: BigNumber;
   pendingRaw: BigNumber;
   balanceFiat: number;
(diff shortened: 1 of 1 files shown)


keerifox (approved):
seems to be working fine

keerifox:
although there seems to be another issue that isn't present on master branch

after receiving a tx,accounts dropdown/accounts page still show blue labels with incoming amounts despite there no longer being any (account.pending doesn't decrease?)

Joohansson:
> although there seems to be another issue that isn't present on master branch
> 
> after receiving a tx,accounts dropdown/accounts page still show blue labels with incoming amounts despite there no longer being any (account.pending doesn't decrease?)

Fix one bug introduces another... Could you please verify now =)

keerifox:
works now!

also seems to have solved the issue where incoming balance card would re-appear, even if there was no change to the incoming balance amount, when finding incoming blocks

PR #168, opened 2020-08-19
Fix "Unable to connect to the server" alert sometimes not displaying

keerifox:
`version` and `confirmation_quorum` requests change the node's state to `loading`, but they provide an argument `skipError` that's set to true, causing all failed requests to keep the state in `loading` rather than change it to `offline`


Joohansson:
Looks legit. I introduced the skipError because some backends do not support version and quorum calls, and would instantly set the node to offline since I request that for the node stats.

PR #174, opened 2020-08-20
Prevent wallet export if the wallet password is not set

### LINKED ISSUE #155: Prevent wallet export when password is blank
Require user to set a secure password before they can export the wallet as url, to provide a security layer above plain text seed

the link could remain in browser history or be shared by accident, the password could at least provide some time for user to move funds to a different seed

keerifox:
fixes #155


Joohansson:
Should we have an additional warning here in red text? For example
"If you forget your password, you will NOT be able to restore your wallet with this link. It's strongly recommended to always store the seed and/or mnemonic for the wallet."

image

I mean in case people use this tool instead of saving the seed.

PR #183, opened 2020-08-22
feat(create-wallet): Sanitize accounts containing bad words

Joohansson:
- When generating a new seed, check the first accounts and redo the seed if any bad words are found
- Delivering clean accounts for the user up to index 19.
- Optional

image

image

Suggested improvement: Replace the for loop and word list with some huge regex for higher speed.


keerifox:
i'm honestly not sure if this should be part of the wallet

one issue is it reduces randomness of generated seeds, especially with 3 character word matches

the other is how it refers to certain words as "bad words", which will make a lot of people rightfully offended
if someone _really wants_ to avoid having words "gay" or "trans" in their address, they can create an extra address

a less minor issue is it introduces complexity to wallet creation, which might result in users spending a lot of time wondering what is a sanitized account and if they need them, instead of learning about the importance of wallet seeds

PR #189, opened 2020-08-24
refactor(desktop-app): Dynamic window size based on screen size

### LINKED ISSUE #166: Desktop app window size
the desktop app starts at 1000px x 600px for me, which is very small. i would give it at least 1300px x 900px on a 1080p screen (which is 67% of screen width and 83% of screen height, could round them up to 70% and 85% and keep 1000x600px as minimum)

optionally remember the last window size

Joohansson:
70% of width and 85% of height
Fixes #166

### DIFF
--- a/desktop-app/src/desktop-app.ts
+++ b/desktop-app/src/desktop-app.ts
@@ -1,6 +1,6 @@
 import 'babel-polyfill';
 
-import { app, BrowserWindow, shell, Menu, protocol, webFrame, ipcMain } from 'electron';
+import { app, BrowserWindow, shell, Menu, screen } from 'electron';
 import { autoUpdater } from 'electron-updater';
 import * as url from 'url';
 import * as path from 'path';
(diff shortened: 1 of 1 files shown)


keerifox (approved):
not able to test this at the moment but the code looks good

desktop-app/src/desktop-app.ts
```diff
@@ -14,10 +14,19 @@ initialize();
...
 
 function createWindow () {
+  let newWidth = 1000;
+  let newHeight = 600;
+  try {
+    const mainScreen = screen.getPrimaryDisplay();
+    const dimensions = mainScreen.size;
+    newWidth = Math.max(newWidth, Math.round(dimensions.width * 0.7));
+    newHeight = Math.max(newHeight, Math.round(dimensions.height * 0.85));
+  } catch {}
```
agilob, line 24: why is it in try/catch, what if it fails? no fallback values are provided?

Joohansson: @agilob The fallback values are above the try, 1000 and 600. I don’t know if it can fail but I put it like that just in case some OS refuse to give the values.

PR #193, opened 2020-08-25
Hide "Missing accounts" hint in private key mode

keerifox:
clearer message when attempting to add an account in private key mode
`Unable to add new account: Cannot add another account in private key mode`
->
`Unable to add new account: Wallet consists of a single private key.`

adds walletService.isSingleKeyWallet() which might be useful later for any UI tweaks exclusive to single key wallets


Joohansson (approved):
Looks good

PR #195, opened 2020-08-27
feat(wallet-import): Added support for bip39 mnemonic

### LINKED ISSUE #194: Add support for importing a bip39 mnemonic
To not mess with the internal wallet functions (and other Nault features) that are dependent on a nano seed (which cannot be obtained from a bip39 mnemonic), the suggestion is:

- Add a new import type called "Bip39 Mnemonic Phrase" (will support 12, 15, 18, 21 or 24 words)
- Add two additional fields for it: Account Index and Password (optional)
- Internally convert the mnemonic to a private key corresponding to the index and optional password
- Import the wallet as a single private key

Everything else will work as normal. This can be used to recover a Ledger account for example,  or from another multi-currency wallet.

Joohansson:
- 12, 15, 18, 21 or 24 words
- Converted to bip39 seed and then imported as a single private key wallet
- No other Nault features affected

image

Fixes #194 

The use of bip39 password has been verified working using this bip39 fork
https://github.com/iancoleman/bip39/pull/434

### DIFF
--- a/src/app/components/configure-wallet/configure-wallet.component.html
+++ b/src/app/components/configure-wallet/configure-wallet.component.html
@@ -65,22 +65,56 @@
     <div uk-grid *ngIf="selectedImportOption === 'seed'" class="uk-margin-top">
       <div class="uk-width-1-1">
         <p>
-          Enter your 64 character seed from any Nano wallet to import it below.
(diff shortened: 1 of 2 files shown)


keerifox (approved):
seems to work as expected

PR #197, opened 2020-08-27
Fix uptime not being checked for known neutral/avoid representatives

keerifox:
includes uptime discrepancy corrections (when weekly uptime is out of date with daily uptime, or rep is up but weekly uptime is 0%)

includes temporary switch from weekly to monthly uptime (until issues are resolved)

### DIFF
--- a/src/app/services/representative.service.ts
+++ b/src/app/services/representative.service.ts
@@ -165,38 +165,61 @@ export class RepresentativeService {
       }
 
       if (knownRep) {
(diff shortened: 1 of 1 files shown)


Joohansson (approved):

My1:
> includes temporary switch from weekly to monthly uptime

I think monthly would be generally better imo because a rep may be down for a few days e.g. if hardware blasted or whatever, which takes time to fix

PR #200, opened 2020-08-29
feat(wallet-export): All active account indexes instead of last index

Joohansson:
- Allow import of accounts 1,2,7,555
- Before, it would import 0 to 7
- Increased the QR size a little bit because if user have 100 accounts, more info needs to be encoded
- Backward compatible
- Fixes 184

**Old json:**
{"accountsIndex":4,"seed":"xyz"}

**New json:**
{"indexes":[0,1,2,3,500],"seed":"xyz"}

### DIFF
--- a/src/app/components/configure-wallet/configure-wallet.component.ts
+++ b/src/app/components/configure-wallet/configure-wallet.component.ts
@@ -322,7 +322,7 @@ export class ConfigureWalletComponent implements OnInit {
       const fileData = event.target['result'] as string;
       try {
         const importData = JSON.parse(fileData);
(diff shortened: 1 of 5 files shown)


keerifox (approved):
looks good and seems to work correctly

maybe we could remove references to `loadAccountsFromIndex()` and `wallet.accountsIndex` in the wallet service

in `loadStoredWallet()` there's also code for loading an already unlocked wallet which i believe was disabled back in nanovault so we probably don't have to support that as nault never stored password in local storage
```js
// Remove support for unlocked wallet
this.wallet.locked = walletJson.locked;
this.wallet.password = walletJson.password || null;
```

PR #202, opened 2020-08-30
Reconnecting to backend after deleting all data

### LINKED ISSUE #201: Issues after clearing all data
imported a seed after clearing all data (without refreshing the page), resulted in seeing this:

image

claims to be in offline mode despite server backend being set to Random

`ERROR Error: Uncaught (in promise): TypeError: s is undefined scanAccounts`

someone else had a different issue where after clearing data, reportedly upon seed import they would get a page with new seed/mnemonic instead

Joohansson:
Fixes #201 

Deleting all data literally delete all settings as well, which causes trouble when trying to use the backend since there is none. This fix updates the server settings after a reset to get the minimum amount of settings to make the wallet function. So it's more like a reset of the settings to default values than a full delete.

@keerifox I think we need to tune some notifications because now it say all settings are removed from local storage but that's not true if we want to keep some default settings. For example here
https://github.com/Nault/Nault/blob/master/src/app/components/configure-app/configure-app.component.ts#L463
and
https://github.com/Nault/Nault/blob/master/src/app/components/configure-app/configure-app.component.ts#L474
and 
https://github.com/Nault/Nault/blob/master/src/app/components/configure-app/configure-app.component.html#L389

### DIFF
--- a/src/app/components/configure-app/configure-app.component.html
+++ b/src/app/components/configure-app/configure-app.component.html
@@ -386,7 +386,7 @@
             <button class="uk-button uk-button-danger" (click)="clearWalletData()" uk-tooltip title="This will delete all data related to your wallet (Your seed and accounts)">Delete Wallet Data</button>
           </div>
           <div class="uk-width-1-3@m">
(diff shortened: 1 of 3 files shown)


keerifox (approved):
yeah maybe we could mention it resets settings to default rather than completely deletes all local storage data

code seems good

PR #205, opened 2020-09-05
feat(transactions): Removed denominations and extended raw support

### LINKED ISSUE #100: Reducing Nano Denominations and only use NANO
Nault is using NANO, knano and nano.

image

Does anyone EVER use anything other than NANO? I think it may just be confusing and dangerous but some users still want it for a nostalgic purpose I don't know. I don't think any other wallet has this feature. This is an old reference to the original raiblock system where Mray (NANO=1M nano) was chosen as the default denomination when exchanges started to list it.

The suggestion is to remove them and only keep NANO.
Related: #93

### LINKED ISSUE #192: Restrict amount text field input to digits and one period
image

Joohansson:
- Removed knano and nano. Everything is NANO (kept some code references if we want to change in the future)
- Allow sending 1 raw
- Detect bad amount for send screen and remote send block modal
- Display incoming raw
- Display addional raw when sending "max" (also for remote send block modal)

Fixes #100 
Fixes #192

### DIFF
--- a/src/app/components/account-details/account-details.component.html
+++ b/src/app/components/account-details/account-details.component.html
@@ -353,31 +353,15 @@
             <label class="uk-form-label" for="form-horizontal-text">Amount</label>
             <div class="uk-form-controls">
               <div uk-grid>
-                <div class="uk-width-3-5">
+                <div class="uk-width-1-1">
                   <div class="uk-inline uk-width-1-1">
                     <a class="uk-form-icon uk-form-icon-flip" (click)="setMaxAmount()" style="padding-right: 7px;" uk-tooltip title="Set Maximum Amount">Max</a>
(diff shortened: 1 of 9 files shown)


keerifox (approved):

aleksre:
Amount input type should be `number`

Joohansson:
> Amount input type should be `number`

How do you make input type number to allow decimals? I tried set step="0.000000000000000000000000000001" but then anything below 0.00001 is detected as invalid. Plus the tooltip was still showing it's a bad number. I keep it as type text for now.
image

My1:
> > Amount input type should be `number`
> 
> How do you make input type number to allow decimals? I tried set step="0.000000000000000000000000000001" but then anything below 0.00001 is detected as invalid. Plus the tooltip was still showing it's a bad number. I keep it as type text for now.

https://stackoverflow.com/a/34057860/4426048

Issue #148, opened 2020-08-11
account labels style needs improvement


keerifox:
current labels:
![image](https://user-images.githubusercontent.com/29272208/89889350-fcd14b80-dbc0-11ea-97fa-827d19ca1cff.png)

suggested style by ajsharkk:

![image](https://user-images.githubusercontent.com/29272208/89889146-b54abf80-dbc0-11ea-9db3-5d4011e5a034.png)

approximate colors are:

    background: #F1F1F1;
    color: #515151;

border-radius could be around 6px

keerifox:
note: extra vertical spacing between account label and address is good, but should also be added between nano/fiat balances

PR #215, opened 2020-10-17
Improve style of account labels and nano addresses

keerifox:
improved layout and click area of the Accounts page
all nano addresses are now shown in a monospace font (all characters with equal width)
fixed reps not known by MyNanoNinja reported as "offline for the past 0 days"
fixed long account names not being truncated on send/sign confirmation pages
reduced max width of the form elements on the send page
added pencil icon to the address book entries to make it easier to tell the labels can be changed
added a hover hint to incoming balance amount so scrolling down to find out what the asterisk means isn't required


keerifox:
this fixes #148

Joohansson:
Nice update! Tested and looks good. Code ok.

PR #217, opened 2020-10-20
Revamped Ledger support

### LINKED ISSUE #13: Overhaul of the Ledger integration
The wallet currently uses [transport-node-hid](https://github.com/LedgerHQ/ledgerjs/tree/master/packages/hw-transport-node-hid) for electron and [transport-u2f ](https://github.com/LedgerHQ/ledgerjs/tree/master/packages/hw-transport-u2f) for web.

Looking at the table over at [ledger-js](https://github.com/LedgerHQ/ledgerjs) it seems like a good choice but with an annoying window popping up in windows. Maybe it can be improved, maybe not. Maybe [web-usb](https://github.com/LedgerHQ/ledgerjs/tree/master/packages/hw-transport-webusb) can be utilized for better stability, but currently only supported in chrome.

Anyway, if any developer want to give it a try to increase stability, feel free! I will probably not touch this.

### LINKED ISSUE #145: Ledger: Use WebUSB and Increase timeouts for verification
apparently the site loses connectivity to the ledger while verification is happening, which kinda doesnt seem good because you are made to not properly verify your transactions.

this might or might not be a side effect of Nault still using U2F.

Also on Windows 10 this leads to annoying popups from Windows all the time, which is not fun.

switching over to webUSB should help though

https://github.com/LedgerHQ/ledgerjs

aleksre:
In addition to the old U2F transport, Nault (for web) now supports 3 additional transports: WebHID, WebUSB & WebBluetooth. Nault will automatically detect the best transport of the 3 USB-modes supported by the browser and OS and use it when communicating with the ledger. Thus, there are no more annoying popups or short timeouts on modern browsers: Edge, Chrome, Opera, Brave, Android Chrome & Samsung Internet. And virtually all OSes (except iOS).

The detection is based on actual browser support and not heuristics, which means that if Firefox or Safari suddenly start supporting Bluetooth or WebUSB, it will start working automatically with no further code changes.

For Nault Desktop USB works well already, but I have added Bluetooth support here too with Node-BLE. Unfortunately I'm unable to build the desktop version for some reason so this has been coded blind.

The [Ledger Github library](https://github.com/LedgerHQ/ledgerjs) reports stability issues with WebUSB on Windows. I have *NOT* been able to reproduce any stability issues with WebHID nor WebUSB. They seem to function exactly the same from my testing, so not sure if we should prefer WebHID on Windows or not? I had this on initially, but currently WebUSB is preferred on all OSes.

Known Issues:
When the ledger is locked, it is reported as disconnected by Nault. Upon unlocking it seems to still not be able to connect and an exception is thrown in the ledger library. The fix is fairly simple though, just close and reopen the Nano app on the Ledger itself. I have added this tip to the error notifications so all in all it is not a big deal.

PS: I don't have a Ledger X so the Bluetooth support is untested. 

Closes #145 and closes #13


Joohansson:
Amazing revamp @aleksre!
I will test carefully both web and desktop as soon as possible. Don't have a Ledger X either but will try to find someone who does.

aleksre:
Thanks. I downloaded the desktop release that was built by github, and unfortunately the bluetooth support crashes deep within the ledger library. Oh well. I'll have a look at it later, but we may just have to disable bluetooth support for the desktop release. Node-ble is labeled as experimental after all.

desktop-app/src/lib/ledger.ts
```diff
@@ -216,7 +217,7 @@ let sendingWindow = null;
 
 // Create a copy of the ledger service and register listeners with the browser window
 export function initialize() {
-  console.log("Ledger service inializing")
+  console.log('Ledger service inializing');
```
keerifox: seems to be an old typo, can we fix it? ^^

```suggestion
  console.log('Ledger service initializing');
```

src/app/services/ledger.service.ts
```diff
@@ -309,8 +335,9 @@ export class LedgerService {
           this.resetLedger();
         }
         if (!hideNotifications && !resolved) {
-          this.notifications.sendWarning(`Ledger device locked.  Unlock and open the Nano application`);
+          this.notifications.sendWarning(`Unable to connect to the Ledger device.  Make sure your ledger is unlocked.  Restart the Nano App on your ledger if the error persists`);
```
keerifox: also this is minor but i think we should use capitalized word Ledger so it's clear it refers to the device and not the collection of Nano account chains

```suggestion
          this.notifications.sendWarning(`Unable to connect to the Ledger device.  Make sure your Ledger is unlocked.  Restart the Nano App on your Ledger if the error persists`);
```

src/app/components/configure-wallet/configure-wallet.component.html
```diff
@@ -121,12 +121,13 @@ <h2 class="uk-heading-divider">Configure Wallet</h2>
...
         <p>
-          Make sure you have the Nano app installed, then connect your ledger device via USB. <a href="https://docs.nault.cc/2020/08/04/ledger-guide.html" target="_blank" rel="noopener noreferrer">Ledger/Nault User Guide</a><br>
+          Make sure you have the Nano app installed and running on your Ledger, then click one of the buttons below to connect via either USB or Bluetooth (Nano X only). 
+          <a href="https://docs.nault.cc/2020/08/04/ledger-guide.html" target="_blank" rel="noopener noreferrer">Ledger/Nault User Guide</a><br>
           <br>
         </p>
         <div *ngIf="ledger.status === ledgerStatus.NOT_CONNECTED" class="uk-alert uk-alert-warning">
           <b>Ledger Device Not Connected</b><br>
-          Connect your ledger device via USB and make sure you have the Nano app installed, then press the button below to retry
+          Connect your ledger device via USB or Bluetooth and make sure you have the Nano app installed and running, then press the button below
```
keerifox:
```suggestion
          Connect your Ledger device via USB or Bluetooth, and make sure you have the Nano app installed and running, then press the button below
```

src/app/components/configure-wallet/configure-wallet.component.ts
```diff
@@ -113,20 +107,34 @@ export class ConfigureWalletComponent implements OnInit {
...
     await this.ledgerService.loadLedger(true);
     this.notifications.removeNotification('ledger-status');
 
     if (this.ledger.status === LedgerStatus.NOT_CONNECTED) {
-      return this.notifications.sendWarning(`No ledger device detected, make sure it is connected and you are using Chrome/Opera`);
+      return this.notifications.sendWarning(`No ledger device detected.  Make sure the Nano app is running on the Ledger.  Restart the Nano App if the error persists`);
     }
 
     if (this.ledger.status === LedgerStatus.LOCKED) {
       return this.notifications.sendWarning(`Unlock your ledger device and open the Nano app to continue`);
```
keerifox:
```suggestion
      return this.notifications.sendWarning(`Unlock your Ledger device and open the Nano app to continue`);
```

src/app/components/configure-wallet/configure-wallet.component.ts
```diff
@@ -113,20 +107,34 @@ export class ConfigureWalletComponent implements OnInit {
...
+
   async importLedgerWallet(refreshOnly = false) {
     // Determine status of ledger device using ledger service
     this.notifications.sendInfo(`Checking for Ledger device...`, { identifier: 'ledger-status', length: 0 });
     await this.ledgerService.loadLedger(true);
     this.notifications.removeNotification('ledger-status');
 
     if (this.ledger.status === LedgerStatus.NOT_CONNECTED) {
-      return this.notifications.sendWarning(`No ledger device detected, make sure it is connected and you are using Chrome/Opera`);
+      return this.notifications.sendWarning(`No ledger device detected.  Make sure the Nano app is running on the Ledger.  Restart the Nano App if the error persists`);
```
keerifox: think that's the last occurrence of it

```suggestion
      return this.notifications.sendWarning(`No Ledger device detected.  Make sure the Nano app is running on the Ledger.  Restart the Nano App if the error persists`);
```

src/app/services/ledger.service.ts
aleksre: Agreed

Joohansson (commented):
So I tested on web and locally compiled desktop on win10 and Linux. It works fine with USB and ledger S. I also got help testing ledger X via USB to an android phone and that now works as well that didn't work at all before. Same person also tested ledger X over Bluetooth to his phone and while a confirmation dialog pop up to pair the device, Nault did not detect the ledger.

Which means I think Bluetooth should be disabled for desktop (since it crashes) and also on web until it can be confirmed working. Or maybe leave it in the web and have more people testing. At least it doesn't crash.

aleksre:
Interesting. Did any error message pop up? I assume this was also tested on Android. Is it possible to test on a PC with Chrome and the console open? That might give a clue as to what is wrong.

Joohansson:
> Interesting. Did any error message pop up? I assume this was also tested on Android. Is it possible to test on a PC with Chrome and the console open? That might give a clue as to what is wrong.

Android yes. The only error was in Nault that it couldn't detect it. And it kept opening up ledger live instead even when it was closed according to the tester. He can probably test on PC coming days. I'll await the results.

aleksre:
If he got 'Unable to connect to the Ledger device.', It means it reached the timeout without getting the config, or that the config failed on line 324. 

Config should only fail if the wrong app is open according to docs. There is a chance the nano app does not support bluetooth at all though, since we are probably the first to test this. The tester should at least make sure he has the latest version of the nano app installed.

It is also possible it just took longer than 2,5 seconds to connect. I'll increase the timeout just in case. But console will give more info.

Joohansson:
Got this from the tester below. Still no luck but I would say that we release this now anyway and possibly solve later once one of us have a Ledger X to debug. This PR makes it much better than before and everthing that worked before works now too. 

” Trying it on a chromebook and still hitting issues.  Stuck at "Checking for Ledger Device" after selecting it for pairing and confirming the pairing request ”

PR #219, opened 2020-11-25
Implemented ledger usb legacy button

Joohansson:
While troubleshooting browsers on windows, my ledger stopped working with webUSB. Added a button that forces legacy U2F protocol, which still works. There is a high risk other users may experience this too, at least one I know so far.

**Windows**
Chrome (v87): OK
Brave (v1.17.73, chromium v87): OK
Firefox: Works but popup keeps popping up for each wallet action like loading accounts and sending. Had to press reconnect after it was connected to keep going.
Edge Chromium based (v87): OK

**Ubuntu 20**
Chrome (85.0.4183.102): OK
Brave (Version 1.9.76 Chromium: 81.0.4044.138): OK
Firefox (82.0): OK but a popup on every communication like on windows. Using legacy U2F protocol. After importing the accounts I had to press reconnect but everything seems to work.

**Problems on Windows:**
Got "Device not detected" after testing a few different browsers. Restarting browser and ledger (and ledger nano app) didn't work.
Starting incognito tab in Chrome: Get the external window pop up asking to connect and "Nano S" is in the list (the name was some random number before when it worked). Clicking connect just give the same message. Everything still works in Firefox using U2F and desktop app. Also Ubuntu no problem.

```
Ledger Device Not Connected
Connect your Ledger device via USB or Bluetooth, and make sure you have the Nano app installed and running, then press the button below
```

This is where it fails
https://github.com/Nault/Nault/blob/master/src/app/services/ledger.service.ts#L292

**Workaround:**
Forcing U2F still works in Windows but with the annoying popup as before.
image

**Also tested:**
Restarted the computer, turned off/on the ledger, updated firmware, reinstalled nano app. Changed USB connector. Incognito tab in chrome. Nothing works so far. Something is blocking the protocol on my machine. Some chrome update?


aleksre:
Did you try WebHID? This may be the WebUSB "instability" mentioned on the ledger library github page. If it works we can prefer WebHID on Windows instead.

If that also doesn't work, I'd personally prefer a more "seamless" user experience, maybe we can set a flag for each transport to blacklist it if it fails, so you'll end up with U2F after 2 failed attempts?

For what it's worth, WebUSB still works just fine for me.

Joohansson:
WebHID is not supported on Chrome it seems. It jumps directly to U2F. "this.supportsWebHID" is false.
Could you please come online in Discord instead. It's so much quicker to discuss a solution.
Your approach sounds like a better one if it can be done with a reasonable amount of effort.

Joohansson:
Better PR: https://github.com/Nault/Nault/pull/220

PR #221, opened 2020-11-25
Prefer WebHID on Windows

aleksre:
Changes in this PR:
- Due to apparent stability issues with WebUSB on Windows, we will instead prefer WebHID for Windows only, if supported.
- ~~If one of the USB transports fail, they will be temporarily blacklisted. Clicking Connect by USB again will detect the next-best option. As always, U2F is the last fallback.~~
- Added Nault to WebHID origin trial, which should enable WebHID for all users of Chrome 86 to 88. Once Chrome 89 is released in February 2021, it will be enabled automatically for everyone and the meta tag can be removed. This will only work on Nault.cc, not localhost.

### DIFF
--- a/src/app/services/ledger.service.ts
+++ b/src/app/services/ledger.service.ts
@@ -117,8 +117,14 @@ export class LedgerService {
    * Detect the optimal USB transport protocol for the current browser and OS
    */
   detectUsbTransport() {
-    if (this.supportsWebUSB) {
-      // Prefer WebUSB
+    const isWindows = window.navigator.platform.includes('Win');
+
+    if (isWindows && this.supportsWebHID) {
+      // Prefer WebHID on Windows due to stability issues with WebUSB
+      this.transportMode = 'HID';
+      this.DynamicTransport = TransportHID;
+    } else if (this.supportsWebUSB) {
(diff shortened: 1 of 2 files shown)


keerifox (approved):
code seems ok

Joohansson (commented):
Your PR requires the user to press the connect button several times while my PR will switch the protocol automatically with a single click. And I don’t think your solution will work if the user refreshes the browser while the ledger is connected. That is when they lunch Nault next time, with the Ledger already set up and connected. The user must NOT need to go and import the ledger again.

So you will either need to implement that (preferably also the connect single-button-click since it's more user friendly) or approve my PR instead.

https://github.com/Nault/Nault/pull/220

aleksre:
Yeah the blacklist has some unfortunate side effect when refreshing so I removed it. I don't feel comfortable just auto-retrying with a different transport when it fails though. My suggestion is we go live with the WebHID changes which should hopefully be enough. If it turns out it isn't, we can reevaluate our options.

My1:
sounds definitely useful especially as at least some PCs I played with didnt like WebUSB devices and just fails to recognize them as anything really

PR #224, opened 2020-12-08
Ledger bluetooth reconnect

### LINKED ISSUE #222: Reconnect Ledger X when using bluetooth on browser refresh
Ledger X has been confirmed working using bluetooth and windows 10, using nault.cc.
But when refreshing the browser, the ledger does not connect using BT (I guess it defaults to webHID). The wallet has to be imported again.

 @aleksre An easy fix?

The following tip was made: Make sure the ledger's BT is on and the nano applet runs BEFORE opening nault

aleksre:
Fixed in this PR:
- Nault will save the Ledger connection type to app settings, and set the transport back to bluetooth if this was the last used connection type. Thus the reconnect link in the nav (or any other calls to `loadLedger()`) will always use the correct connection type.
- Updated all Ledger related packages. This actually fixed the broken WebUSB support for Windows. WebHID is still the preferred transport but it's nice to have it working again.
- Bug fix/workaround: Previously, Nault would automatically reconnect the ledger wallet when launched, however this was actually broken because it would attempt to reconnect before the transport support had completed (which is async), causing it to attempt to reconnect using U2F every time. It would be possible to delay this until browser support has been established, but even then the reconnection may prompt a dialog, which the user did not ask for. This is not ideal, and actually disallowed by the ledger library, which requires the that connection is triggered by a click. I have therefore removed this behavior for now. Reconnecting is anyway very easy with a single click on "Click here to reconnect".

Deployed to: https://nault-h8xlpxc5e.vercel.app/

Closes #222


My1:
looks like a confirm for me, obviously this means that switching transport means to set up fresh but better than it not working at all

Joohansson (approved):
Code looks good

PR #226, opened 2020-12-10
Add section for transferring Nano between own accounts

### LINKED ISSUE #191: Send page tab for sending between own accounts
image

no need to duplicate layout/code for this, simply switch between text field and dropdown based on which tab it is

keerifox:
seems to be working as expected, but i would appreciate some testing, to make sure the funds are sent exactly where the user wants them to be sent!

fixes #191


Joohansson (approved):
Tested this on both web and desktop, also with ledger. Works perfectly.
Even when the accounts were not loaded yet after importing a new wallet, they populated correctly in the "own account" list while they were being loaded. Nice work!

PR #239, opened 2021-01-04
fix: Proper validation of seed, mnemonic and private key during import

### LINKED ISSUE #238: Importing seed allows you to enter a non hex character (A-F 0-9)
Must implement a check here because it allows seed such as zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz

Not sure how that even works but it's very bad since the mnemonic does not seem to be created.

Joohansson:
Fixes #238
To block importing of invalid characters


BitDesert (approved):
Confirmed working.

PR #242, opened 2021-01-04
Ledger type

### LINKED ISSUE #231: Ledger Nano S/X - Which connection type is supported?
When configuring a new wallet with a Ledger device it tells you to "click one of the two buttons" and connect via Bluetooth even if your browser/app doesn't support Bluetooth.

Maybe we could tell the user which options are available, maybe with labels:
grafik

We already have the information on which options are available `ledgerService.supportsBluetooth`:
https://github.com/Nault/Nault/blob/138c7ccfa286ebd6f310990a7a6578994591a5a6/src/app/components/configure-wallet/configure-wallet.component.html#L174

BitDesert:
Fixes #231 

The "New Ledger Wallet" view now looks like this so it's easier to identify:
grafik


Joohansson (approved):

Joohansson:
Looks better

PR #255, opened 2021-01-11
Wallet unlock form improvements

keerifox:
password input and unlock button are now disabled for 0.5s upon unsuccessful unlock attempt, rather than delaying any unlocking attempt by 0.5s (even if the password was entered correctly from the first try, the unlocking process used to be delayed)

image

also changes wording from "Invalid password" to "Incorrect password"

### DIFF
--- a/src/app/components/wallet-widget/wallet-widget.component.ts
+++ b/src/app/components/wallet-widget/wallet-widget.component.ts
@@ -1,4 +1,4 @@
-import { Component, OnInit } from '@angular/core';
+import { Component, ElementRef, OnInit, ViewChild } from '@angular/core';
 import {WalletService} from '../../services/wallet.service';
 import {NotificationService} from '../../services/notification.service';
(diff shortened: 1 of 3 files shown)


src/app/components/wallet-widget/wallet-widget.component.ts
```diff
@@ -67,7 +69,24 @@ export class WalletWidgetComponent implements OnInit {
   }
 
   async unlockWallet() {
-    await new Promise(resolve => setTimeout(resolve, 500)); // brute force delay
+    if (this.shouldDelayUnlockAttempt === true) {
+      await new Promise(resolve => setTimeout(resolve, 500)); // brute force delay
```
Joohansson: I think some additional logic must be set here. Since the function is async you can press the button and bruteforce how much you want. The delay has no affect? Maybe set a global variable that is reset after 500ms. And if that variable is set, just return the unlockWallet function. I made this wrong from the very beginning it seems.

PR #257, opened 2021-01-17
Ledger stability and fixes

### LINKED ISSUE #234: setDebugMode is deprecated for ledger devices?
Hi,

Got the following message when doing some transactions using version 1.8.6:

```setDebugMode is deprecated. use @ledgerhq/logs instead. No logs are emitted in this anymore.```

### PR DESCRIPTION
**Pull request by aleksre**
Fixes in this PR:
- Improved Ledger stability. There were several situations where the Ledger would not be reset properly after a disconnect, being locked, or unsuccessful connect. This would cause it to fail on consecutive connection attempts unless the user closed and reopened the Nano app, or did a browser refresh. This should no longer be necessary. It's possible some of the reports of the Ledger connection only working in incognito mode were actually caused by this, but I don't know that for sure. In any case, it should feel a lot more stable.
- Cleaned up Ledger related notifications:
  - Will no longer display connection error notification when the user intentionally dismisses the connection prompt. 
  - Fixed an issue when both the "Ledger device locked" and "Successfully connected to Ledger device" would show at the same time. 
  - Fixed an issue where two simultaneous error notifications would be displayed when failing to reconnect from the side bar.
- Changed the text in the side bar when the Ledger is locked. Before it would say the app was closed which is not accurate.
- Removed deprecated `.setDebugMode(true)` and replaced it with @ledgerhq/logs on both desktop and web. ~~Didn't see anything get logged yet though.~~ Actually it is spamming.
- Safari is apparently completely unsupported. Doesn't even support U2F. Added a check to see if any USB mode is supported and if not, the USB button is disabled and we show a unsupported message: 

image

Closes #234

### CODE REVIEW

**Review by keerifox — APPROVED** [review]

**Comment by aleksre** [comment]
Spoke too soon on a few issues. LedgerLogs is working, and is borderline spamming the console, so maybe we should remove or comment out that line and enable it only when needed?

image

I did most of the testing on Ledger X, but the S is still "crashing" when it becomes locked after a few minutes. Nano app still has to be restarted after that before it will reconnect. I wonder if this is a bug with the App on the ledger itself. X works well.

**Edit:** Just tried it again on the S, and this time it worked! So apparently its random or cable related 🤷 

When it fails, this is the error:
```
// ledger.service.ts (line 365):
App config error:
TransportStatusError {
  message: "Ledger device: Internal error, please report (0x6faa)"
  name: "TransportStatusError"
  stack: "Error↵    at new TransportStatusError (http://localhost:4200/vendor.js:58288:18)↵    at TransportWebUSB.Transport.send (http://localhost:4200/vendor.js:71232:15)"
  statusCode: 28586
  statusText: "HALTED"
}
```

Joohansson:
@aleksre I had tons of issues when using the Ledger with a bad USB port. Could be that. I agree to deactivate the logging if it's spamming too much. Maybe enable it somehow via the package.json and "npm run wallet:dev-debug" or something. Not sure.

aleksre:
Yeah I agree. I just commented it out for now.

Joohansson (approved):
Looks good to me. Also tested my Ledger

BitDesert (approved):
Looks good! 
I'm fine with letting it being commented out for now.

PR #258, opened 2021-01-18
Localization

BitDesert:
Using ngx-translate and not the native Angular translation library so we don't compile the project for every language.
As the default it uses the browser language and falls back to english if the language or string is not available.

When a new language gets added it gets used if you don't force a language in the settings.


BitDesert:
This branch differs too much (closing in favor of #457 )

PR #262, opened 2021-01-20
New rep summary page

### LINKED ISSUE #188: Misleading message on the Representatives page
the message "None of your accounts have delegated any voting weight yet" is displayed while the representatives list is loading, even if the user has accounts with weight delegated to reps

keerifox:
before:

image

after:

image

i wasn't able to show account indexes here e.g. `Account #7`, as the index isn't part of the account object. perhaps someone could implement that later on

\-

reps book also undertook some changes, mostly to provide clarity and avoid possible confusion with the list of used reps

before:

image

after:

image

\-

fixes #188

### DIFF
--- a/src/app/components/account-details/account-details.component.html
+++ b/src/app/components/account-details/account-details.component.html
@@ -130,7 +130,7 @@
                     <div *ngIf="!showEditRepresentative">
                       <div uk-grid>
                         <div class="nano-address-monospace uk-width-auto uk-text-truncate" style="max-width: calc(100% - 35px);">
-                          <span *ngIf="repLabel" class="uk-label uk-label-danger">{{ repLabel }}</span> 
+                          <span *ngIf="repLabel" class="account-label rep">{{ repLabel }}</span> 
                           <app-nano-account-id [accountID]="!account ? '':account.representative"></app-nano-account-id>
                         </div>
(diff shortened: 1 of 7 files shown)


Joohansson (approved):
Nice work, looks good!

BitDesert (changes requested):
I don't know if the labels are a bit too flashy when hovering them.
They change colors often and the change is not consistent (rep changes from purple to grey with purple border, account changes from grey to solid blue).

Additionally it would be great when the list below would reset when you click on a different account/rep. Currently it just adds it to the list.

keerifox:
> I don't know if the labels are a bit too flashy when hovering them.
> They change colors often and the change is not consistent (rep changes from purple to grey with purple border, account changes from grey to solid blue).

this is intentional, accounts and reps have their own distinct styles and are presented as is to be consistent with how they look throughout the rest of the application and to accent color only on representatives rather than reps and accounts both


> Additionally it would be great when the list below would reset when you click on a different account/rep. Currently it just adds it to the list.

that's the old behavior: you click on everything that you wish to change, and it is added to the list allowing you to change several reps or accounts at once

BitDesert (approved):
Alright, thanks @keerifox !

Issue #256, opened 2021-01-14
Incoming transactions sometimes not being processed


Joohansson:
There have been reports of incoming transactions that are simply not being processed unless you clear the cache from settings or run Nault in an incognito tab.

Not clear in which version this happened and have not yet been able to reproduce. Maybe it's already solved. I'll leave this issue open for now in case someone else gets stuck.

PR #267, opened 2021-01-21
Improving the PoW threshold used for different block types

### LINKED ISSUE #264: Generate lower difficulty PoW for open/receive blocks
most backends still do not support lower difficulty, so this would only affect client-side PoW generation in the short term

right now there is just one cache:
1. work cache for both sends and receives (higher difficulty)

proposed:
1. work cache for both sends and receives (higher difficulty)
2. work cache for receives only (lower difficulty)

when looking for lower difficulty PoW both caches are checked in this order: higher, lower (for network prioritization)

\-

precache remains at higher difficulty to avoid unnecessary requests/computation

Joohansson:
A first try to reduce the work needed for open/receive blocks.

Problem is Nault is not designed to support this from the beginning and the way it always tries to cache work in advance makes it more or less impossible to solve. When you precache work you don't know what the next block will be so the only time you can take advantage of lower pow is when requesting a receive block with no existing cache.

**There is only one use-case found when that happens (all of these must be fulfilled):**
* User has Nault open with no existing work cache
* A transaction is incoming or has been sent earlier but not yet received (for example when using manual receive)
* Opens the receive screen: Now it will check for pending and cache 1/64 pow since it's most likely that will be the next block

The suggestion from issue 264 was to add 1/64 when a block was requested from the block service here: https://github.com/Nault/Nault/compare/correct-receive-pow?expand=1#diff-19f3dac78ed255aab575eeaf33008b09a262bd8de14b9133294bc771ef68fd24R300

But I can't find a single occurrence when that happens AND there is no cache. Either the 1x work cache has already been added via the [reloadBalances() function](https://github.com/Nault/Nault/compare/correct-receive-pow?expand=1#diff-9346ac0e59e5e9a58990fc61941ecfc327b8584c175198647a67d8536bef9a00R764) (which is called all over the place) or it has been added after a previous receive action here: https://github.com/Nault/Nault/compare/correct-receive-pow?expand=1#diff-19f3dac78ed255aab575eeaf33008b09a262bd8de14b9133294bc771ef68fd24R305

Which means, even when receiving many incoming at once, it will still use 1x pow.

**EDIT**
Now it skips adding new work directly after a receive. It will check for pending and if any pending using 1/64 instead. If there are no more pending, it will use 1x. That means multiple receive (auto or manually) will use the lower pow now (except the first that will use whatever in the cache). The pow API service has been updated to include difficulty as parameter. This will be utilized by bpow backends (proxy.nanos.cc) but not yet dpow.

fixes #264

### DIFF
--- a/src/app/components/receive/receive.component.ts
+++ b/src/app/components/receive/receive.component.ts
@@ -77,7 +77,12 @@ export class ReceiveComponent implements OnInit {
           if (!frontiers.frontiers.hasOwnProperty(account)) {
             continue;
           }
-          this.workPool.addWorkToCache(frontiers.frontiers[account]);
+          // Technically should be 1/64 multiplier here but since we don't know if the pending will be received before
+          // a send or change block is made it's safer to use 1x PoW threshold to be sure the cache will work.
+          // On the other hand, it may be more efficient to use 1/64 and simply let the work cache rework in case a send is made instead
+          // The typical user scenario would be to let the wallet auto receive first
+          console.log('Adding pending to work cache');
+          this.workPool.addWorkToCache(frontiers.frontiers[account], 1 / 64);
         }
       }
     }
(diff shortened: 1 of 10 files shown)


keerifox:
https://github.com/keerifox/Nault/commit/4a5e34e51642813d5570bab2584ae2bec9140d8c
https://github.com/keerifox/Nault/commit/776370fe1de4c47ccc4f9bb46816f2104f070510

simplified work pool code. fixes at least one of the causes for issue #256

keerifox (approved):

PR #276, opened 2021-01-25
Allow custom local pow difficulty

### LINKED ISSUE #273: Add an option to specify multiplier when using local PoW
To let the user prioritize transactions. Only for CPU, GPU, and possibly future p2PoW

Joohansson:
- Only for CPU / GPU pow
- Overrides default SEND & RECEIVE (even if receive is 1/64x)
- Display pow notification until work is finished
- Can take VERY long time above 4x (GPU pow possibly never finish when 32x for example?)
- Resets local work cache when increasing multiplier (if user approve)

Fixes #273 

image

image

### DIFF
--- a/src/app/components/configure-app/configure-app.component.ts
+++ b/src/app/components/configure-app/configure-app.component.ts
@@ -112,6 +112,7 @@ export class ConfigureAppComponent implements OnInit {
 
   multiplierOptions = [
     { name: 'Default (1x or 1/64x)', value: 1 },
+    { name: 'Automatic (max 8x)', value: 0 },
     { name: '2x', value: 2 },
     { name: '4x', value: 4 },
     { name: '8x', value: 8 },
--- a/src/app/components/configure-app/configure-app.component.ts
+++ b/src/app/components/configure-app/configure-app.component.ts
@@ -104,11 +104,22 @@ export class ConfigureAppComponent implements OnInit {
 
   powOptions = [
     { name: 'Best Option Available', value: 'best' },
-    { name: 'Client Side - WebGL [Recommended] (Chrome/Firefox)', value: 'clientWebGL' },
-    { name: 'Client Side - CPU', value: 'clientCPU' },
-    { name: 'Server - Nault Server', value: 'server' },
+    { name: 'Client Side - GPU/WebGL', value: 'clientWebGL' },
+    { name: 'Client Side - CPU (Slowest)', value: 'clientCPU' },
+    { name: 'Selected Server - dPoW / boomPoW / Other', value: 'server' },
   ];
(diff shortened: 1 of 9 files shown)


keerifox (approved):
seems to be working fine, code also looks ok

src/app/components/configure-app/configure-app.component.ts
```diff
@@ -104,12 +104,24 @@ export class ConfigureAppComponent implements OnInit {
 
   powOptions = [
     { name: 'Best Option Available', value: 'best' },
-    { name: 'Client Side - WebGL [Recommended] (Chrome/Firefox)', value: 'clientWebGL' },
-    { name: 'Client Side - CPU', value: 'clientCPU' },
-    { name: 'Server - Nault Server', value: 'server' },
+    { name: 'Client Side - GPU/WebGL', value: 'clientWebGL' },
+    { name: 'Client Side - CPU (Slowest)', value: 'clientCPU' },
+    { name: 'Selected Server - dPoW / boomPoW / Other', value: 'server' },
```
BitDesert: For consistency with "Client Side":
```suggestion
    { name: 'Server Side - dPoW / boomPoW / Other', value: 'server' },
```

Are the possible PoW mechanisms relevant "dPoW / boomPoW / Other" ?

src/app/components/configure-app/configure-app.component.ts
```diff
@@ -300,12 +316,30 @@ export class ConfigureAppComponent implements OnInit {
...
+      if (await this.clearWorkCache()) {
+        reloadPending = true; // force reload balance => re-work pow
+      }
     }
 
     const newSettings = {
       walletStore: newStorage,
       lockInactivityMinutes: Number(this.selectedInactivityMinutes),
       powSource: newPoW,
+      multiplierSource: Number(this.selectedMultiplierOption),
```
BitDesert, line 342: Why do we need to cast `this.selectedMultiplierOption` to Number? Isn't it already set to number in line 123?

BitDesert (changes requested):

src/app/components/configure-app/configure-app.component.ts
Joohansson: Yeah that was strange. Without the Number, it got converted to string in the local store. Casting to Number solved the issue

src/app/components/configure-app/configure-app.component.ts
Joohansson: Solved

keerifox:
small fix for remote PoW option name: https://github.com/keerifox/Nault/commit/018b49c56d2032ea913491bd5d773d411188ebe3

should be good to go otherwise

PR #281, opened 2021-02-01
Account summary redesign

keerifox:
preview:

also in this PR:

* representative name from mynanoninja is now displayed in account details if no custom label is set
* fixed "+1 raw" being squeezed into "+1...1 raw"
* fixed raw amount not being shown when hovering mouse over incoming transaction amounts
* minor style improvements for the offline signing modal
* removed unused modal/code for changing representative from the account details page

### DIFF
--- a/src/app/components/account-details/account-details.component.css
+++ b/src/app/components/account-details/account-details.component.css
@@ -158,6 +158,10 @@
     position: relative;
 }
 
+.representative-details, .balance-details {
+    min-height: 85px;
+}
+
 @media (max-width: 1199px) {
     /* horizontal separator between rep details and balances */
     .representative-details::after {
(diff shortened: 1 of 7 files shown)


BitDesert:
Lookin' good!

Should we apply a `border-radius: 10px;` to `.uk-modal-dialog` ? 
The `.uk-card` has this as well.

keerifox:
> Lookin' good!

thanks!

&nbsp;

> Should we apply a `border-radius: 10px;` to `.uk-modal-dialog` ?

actually yeah. those corners are too sharp

Joohansson:
Design looks good to me. One minor issue is the incoming balance is not updated when you send a new transaction. Must update the page manually or using the refresh button. It's the same in the old one so not sure if it's worth fixing. But I think one reason for that was if you receive 1000 spam transactions, you don't want to update the page 1000 times because it would overload the backend and hit rate limiters.

Does the QR modal work ok on mobile? Not sure how to test it locally.

PR #283, opened 2021-02-04
Added desktop file for Linux

marioortizmanero:
Hi!

I recently downloaded Nault from the [AUR](https://aur.archlinux.org/packages/nault-bin/), which works nicely but doesn't include a desktop file. Rather than updating the [AUR repo](https://github.com/agilob/nault-bin) I thought it'd be more useful upstream so that everyone uses the same desktop file. There's a couple things you might want to modify, like the name or the comment. Also, I don't think the mime type exists yet but I thought it'd be a good idea to make it future proof.


BitDesert:
Do we need anything else or is this just the extra file?
I'm not that familiar with arch linux packages.
I'm asking because of that code that is commented out: https://github.com/agilob/nault-bin/blob/master/PKGBUILD#L25

BitDesert:
I just saw that the latest AppImage already includes a `nault.desktop` file:
https://github.com/Nault/Nault/releases/download/v1.9.4/Nault-1.9.4-Linux.AppImage

marioortizmanero:
Oh! Sorry, I didn't know appimages already included `desktop` files and all. That should work, yes. The only issue I'm facing is that after extracting the appimage the only contents in `usr/share/icons/` are only `hicolor/0x0/apps/nault.png`, where 0x0 doesn't make much sense as a resolution.

Edit: the `Exec` field in the desktop file also seems to be wrong:

```ini
[Desktop Entry]
Name=Nault
Exec=AppRun
Terminal=false
Type=Application
Icon=nault
StartupWMClass=Nault
X-AppImage-Version=1.0.0
Comment=Wallet for interacting with Nano
MimeType=x-scheme-handler/nano;
Categories=Office;Finance;
```

BitDesert:
The current `.desktop` file get's probably generated by `electron-builder`.
Maybe some options are missing here: https://github.com/Nault/Nault/blob/master/package.json#L129
https://www.electron.build/configuration/linux

marioortizmanero:
I'll try that out. I'm new to the project though, how can I build the appimage? There are no instructions in https://github.com/Nault/Nault/blob/master/DEVELOPMENT.md (AFAIK)

BitDesert:
Please refer to the "desktop builds" section: https://github.com/Nault/Nault/blob/master/DEVELOPMENT.md#desktop-builds

Running `npm run desktop:build-local` under Linux should result in your AppImage file at `desktop-app\build`.

marioortizmanero:
OK I figured it out! All that I needed to fix was the generic name and adding the icons, which I converted to the required sizes & filenames with `imagemagick`. Nault can be [packaged for Linux](https://github.com/agilob/nault-bin/pull/1) easily now (after the next release with this change).

Joohansson:
Don't know how to verify but code change looks good to me.

BitDesert (approved):

BitDesert:
Looks good @marioortizmanero , thanks for your contribution!

agilob:
1.9.7 with these changes is now in AUR @marioortizmanero  :)

marioortizmanero:
Great!! 🚀🚀

marioortizmanero:
@BitDesert shall I open a PR indicating that Nault is on the AUR on the README?

BitDesert:
@marioortizmanero Sure, maybe somewhere in the Desktop App section here? https://github.com/Nault/Nault/blob/master/README.md#desktop-app

PR #302, opened 2021-02-14
Include AUR mention

marioortizmanero:
As discussed in #283, this includes a mention of the AUR package for Arch Linux users, if you consider the  wording/placement appropiate.


BitDesert:
Thanks for the PR! Included a `(unofficial)` tag as it's not maintained by us.

marioortizmanero:
Sure!

PR #308, opened 2021-02-15
Alert on long PoW and allow cancellation

### LINKED ISSUE #277: Allow local PoW to be cancelled by the user if it takes too long
* CPU pow should be possible by using worker.terminate() on the worker list
* GPU pow: Use the progressCallback to abort https://github.com/numtel/nano-webgl-pow

Joohansson:
* If local PoW takes longer than 60sec, alert in the sidebar
* Link to PoW methods or let user cancel
* Disabled CPU pow fallback on failed server PoW. Automatic fallback could lead to excessive CPU usage on both desktop and mobile devices. For example, if you have 20 accounts, it could take forever and drain the mobile battery. I think it's better to let the user change pow method if needed via the warning notification. Then the user won't end up waiting forever on pow.
* Changing pow method while local work is ongoing will be cancelled, and remade using new method
* Server pow or custom pow server is currently not monitored regarding long process time. It may be possible to cancel HTTP requests on demand but it's complicated and couldn't figure it out right now

image

**Typical test scenario - PASS**

1. Create an empty wallet with two accounts (will require two PoW)
2. Select CPU as pow method and 64x multiplier
3. Clear the work cache
4. Reload the whole wallet or go to receive screen and press "find incoming" -> wait 60sec -> cancel work in the alert. It should cancel any work in the pipeline with a warning
5. Go to receive screen and press "find incoming". That should trigger pow calculation again. The expected behaviour is that it will retry calculate both pow again. Repeat 4 and 5 or select another pow method.

Fixes #277

### DIFF
--- a/src/app/components/configure-app/configure-app.component.ts
+++ b/src/app/components/configure-app/configure-app.component.ts
@@ -325,6 +325,13 @@ export class ConfigureAppComponent implements OnInit {
       if (newPoW !== 'clientWebGL' && newPoW !== 'clientCPU' && newPoW !== 'custom') {
         this.selectedMultiplierOption = this.multiplierOptions[0].value;
       }
+      // Cancel ongoing PoW if the old method was local PoW
+      if (this.appSettings.settings.powSource === 'clientWebGL' || this.appSettings.settings.powSource === 'clientCPU') {
+        // Check if work is ongoing, and cancel it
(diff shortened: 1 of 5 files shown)


keerifox (approved):

BitDesert (approved):
Checked working fine.

PR #330, opened 2021-02-21
Receive amount in fiat

aleksre:
This PR adds a fiat field to the receive screen so both nano and fiat amounts can be entered, making Nault more useful as a POS device. The fields are now also validated.

It also tweaks the UI to be in line with the other pages like the send page. The QR code is bigger and displays an animation on successful payment that makes it easy for both the merchant and the customer to see that payment was successful. This animation is **only** shown if the amount is the same or above specified. So if the customer were to lower the amount in their wallet before sending it will not be shown.

image
[Old screenshot for comparison](https://imgur.com/HlKqJ7T)

https://user-images.githubusercontent.com/1097444/108627947-595ad080-7458-11eb-8a7c-9c17186cb85b.mp4

### DIFF
--- a/src/app/components/account-details/account-details.component.css
+++ b/src/app/components/account-details/account-details.component.css
@@ -55,12 +55,6 @@
     bottom: 3px;
 }
 
-#refresh {
-    cursor: pointer;
-    margin-left: 10px;
-    position: relative;
-    bottom: 3px;
-}
 .refresh-disabled {
     color: #bbb;
 }
--- a/src/app/components/account-details/account-details.component.html
+++ b/src/app/components/account-details/account-details.component.html
@@ -169,8 +169,11 @@
 
     </div>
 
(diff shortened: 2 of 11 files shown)


aleksre:
Keypad input

image

keerifox:
*disregard, seems to be unrelated to pr bug*

keerifox:
slight tweaks/improvements:

https://github.com/keerifox/Nault/commit/f82099f3aae814bd64dfcfceece60e088159b830

https://github.com/keerifox/Nault/commit/eaf7a685e4280f6201b3300ce0f6dcf7c91fac27

https://github.com/keerifox/Nault/commit/84c294e7625eab9671b4c407b3235030ba3f9efd

the rest seems to be fine except one thing: i'm getting this misleading tooltip on hover

image

keerifox (approved):

PR #342, opened 2021-03-04
Make deeplinks open the send tab automatically on desktop

otonashixav:
Per the title, Nault desktop opens and the send tab is autofilled with address/amount when clicking on deeplinks. Additionally, limited the Nault desktop app to only one instance so that clicking on a deeplink while Nault is already open doesn't open another instance.

I noticed that the QR scan component also has code to handle deeplinks, should this and the added deeplink-related code be moved into a deeplink service? If so, should all the protocols specified by the QR scan component also work on click?

### DIFF
--- a/desktop-app/src/desktop-app.ts
+++ b/desktop-app/src/desktop-app.ts
@@ -13,6 +13,7 @@ const log = require('electron-log');
 let showUpdateErrors = false;
 let saveTimeout = null;
 let isDownloading = false;
+const nano_schemes = ['nano', 'nanorep', 'nanoseed', 'nanokey', 'nanosign', 'nanoprocess'];
 
 /**
  * By default, the logger writes logs to the following locations:
@@ -156,10 +157,10 @@ new AppUpdater();
 
 // Register handler for nano: links
 if (process.platform === 'darwin') {
-  app.setAsDefaultProtocolClient('nano');
+  nano_schemes.forEach((scheme) => app.setAsDefaultProtocolClient(scheme));
 } else {
   const args = process.argv[1] ? [path.resolve(process.argv[1])] : [];
-  app.setAsDefaultProtocolClient('nano', process.execPath, args);
+  nano_schemes.forEach((scheme) => app.setAsDefaultProtocolClient(scheme, process.execPath, args));
 }
 
 // Initialize Ledger device detection
@@ -240,8 +241,7 @@ if (!appLock) {
 
     // on windows, handle deep links on launch
     if (process.platform === 'win32') {
-      const deeplink = process.argv.find((s) => s.startsWith('nano:'));
-
+      const deeplink = findDeeplink(process.argv);
       if (deeplink) handleDeeplink(deeplink);
     }
 
@@ -255,8 +255,7 @@ if (!appLock) {
 
       // Detect on windows when the application has been loaded using a nano: link, send it to the wallet to load
       if (process.platform === 'win32') {
-        const deeplink = argv.find((s) => s.startsWith('nano:'));
-
+        const deeplink = findDeeplink(argv);
         if (deeplink) handleDeeplink(deeplink);
       }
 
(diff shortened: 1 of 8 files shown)


Joohansson:
Nice update. It works fine for me on windows when Nault is not running. The deeplink click opens Nault and navigates to the send screen. It does not navigate to the send screen if Nault is open which is good because you don't want to lose whatever you are doing currently. So when Nault is running in the background and you click on a deeplink it starts to flash in the taskbar but it does not bring the window to foreground, is that possible to solve?

Not sure about the last part with QR reader. Isn't it good as it is? You can only scan if Nault is running and then it works as it should. Or do you mean it should also be possible to use deeplink by clicking on something that requests "nanosign:abc" and similar? It's not implemented anywhere yet but I guess it wouldn't hurt. Maybe do that in a separate PR as it's more difficult to test?

otonashixav:
> It does not navigate to the send screen if Nault is open which is good because you don't want to lose whatever you are doing currently.

Strange, it does nagivate to the send screen on my end. Since you mentioned that not nagivating would be better in this context, should I remove the lines meant to navigate while the app is open? 

> Not sure about the last part with QR reader. Isn't it good as it is? You can only scan if Nault is running and then it works as it should. Or do you mean it should also be possible to use deeplink by clicking on something that requests "nanosign:abc" and similar? It's not implemented anywhere yet but I guess it wouldn't hurt. Maybe do that in a separate PR as it's more difficult to test?

It is good as it is, though I'm reimplementing some functionality that the QR scanner already implemented, and was wondering if it would be better to avoid repitition by creating a service to navigate based on deeplinks and having both the click and QR scanner utilize that service. I'm not sure whether the other protocols should need to work with clicks. 
https://github.com/otonashixav/Nault/blob/e2afa0a3bce4a826a84a223fb107c0efac481706/src/app/components/qr-scan/qr-scan.component.ts#L57
https://github.com/otonashixav/Nault/blob/e2afa0a3bce4a826a84a223fb107c0efac481706/src/app/app.component.ts#L141

Joohansson:
> > It does not navigate to the send screen if Nault is open which is good because you don't want to lose whatever you are doing currently.
> 
> Strange, it does nagivate to the send screen on my end. Since you mentioned that not nagivating would be better in this context, should I remove the lines meant to navigate while the app is open?
> 
Good question. If you do click on a nano link you do expect something to happen, so I think you can keep it but don't have to waste too much time trying to solve it on every machine. Maybe it's my security software that blocks it. As far as I know, there is no way for a website to force a click without user interaction.

> > Not sure about the last part with QR reader. Isn't it good as it is? You can only scan if Nault is running and then it works as it should. Or do you mean it should also be possible to use deeplink by clicking on something that requests "nanosign:abc" and similar? It's not implemented anywhere yet but I guess it wouldn't hurt. Maybe do that in a separate PR as it's more difficult to test?
> 
> It is good as it is, though I'm reimplementing some functionality that the QR scanner already implemented, and was wondering if it would be better to avoid repitition by creating a service to navigate based on deeplinks and having both the click and QR scanner utilize that service. I'm not sure whether the other protocols should need to work with clicks.
> https://github.com/otonashixav/Nault/blob/e2afa0a3bce4a826a84a223fb107c0efac481706/src/app/components/qr-scan/qr-scan.component.ts#L57
> https://github.com/otonashixav/Nault/blob/e2afa0a3bce4a826a84a223fb107c0efac481706/src/app/app.component.ts#L141

If you can make it use less repetition it's a bonus. But it's not a requirement that it must be able to react on clicks for every type. If you want to solve it, then it would be fine. If not, also fine =)

otonashixav:
> If you can make it use less repetition it's a bonus. But it's not a requirement that it must be able to react on clicks for every type. If you want to solve it, then it would be fine. If not, also fine =)

I'll go ahead and do it then, should make NFC easier to implement as well.

Joohansson:
I tested all different QR URL schemes and works fine. However, deeplink clicks will only work with nano: because an electron app can only register one scheme, as far as I know, currently supporting "nano:"

I don't think there is anyway around that. https://github.com/Nault/Nault/blob/master/desktop-app/src/desktop-app.ts#L157

Also, the app now links properly to SEND even when open but it takes 5sec? Maybe not possible to fix. Let me know when you are done making commits.

otonashixav:
I'm done making commits now (hopefully); all the nano schemes should work: `nano`, `nanorep`, `nanoseed`, `nanokey`, `nanosign`, `nanoprocess`.

PR #345, opened 2021-03-11
Update design for incoming transactions list, fix unconfirmed balance appearing across the wallet

### LINKED ISSUE #306: Total wallet balance being reset to 0 for a few seconds immediately after receive
not sure if this happens when an account card is selected, or only for total balance

### LINKED ISSUE #344: Unconfirmed but checked incoming balance has confusing UX
**Describe the bug**
Pending unconfirmed balances show up in a weird state in the UI. An incoming balance will show up like this:

image

The receive window will show no entries:

image

The balances API call shows the pending unconfirmed balance, the subsequent calls don't see the block as it's unconfirmed. When this happens there should probably be a message explaining what's going on.

**To Reproduce**
Have an account with a pending unconfirmed TX
Open receive

**Expected behavior**
The pending TX should be indicated as uncomfirm

**Screenshots**
See above.

**Desktop App (please complete the following information):**
 - OS: Windows 10
 - Version 1.10.2

**Additional context**
Ongoing network spam /

### LINKED ISSUE #346: Switching account in sidebar affects prepared transaction
if you see "You are about to send X NANO", changing account in sidebar shouldn't change "from" account at that point

keerifox:
Some testing would be nice as this PR removes a bunch of code that *appears* unnecessary, and while it seems redundant there's a possibility that some functionality used to depend on that redundancy.

\-

Incoming transactions list on the Receive page has been updated to the newest design that is used in the account summary. The dropdown now affects the list of incoming transactions, filtering it to a specific destination account. User is no longer informed about how much funds or how many transactions under min. receive they're missing as we cannot calculate it based on unconfirmed balance.

old:

image

new (all accounts selected in the dropdown):

image

new (specific account selected in the dropdown)

image

\-

Fixes #344 - Unconfirmed transactions are no longer included in UI when the minimum receive setting is disabled. Incoming balance now only includes amounts from the 50 blocks (currently any blocks but when V22 nano node is out, it should be blocks sorted by transacted amount, as there is a bug in V21 nano node that breaks the sorting)

Fixes #306 - Balance is no longer reset to 0 while updating, instead a spinner icon appears next to it while it is being updated.

Fixes #346 - Switching account in sidebar no longer affects finalized transaction

Fixes characters being cut off:

\-

Known issues:

**[Should be fixed now]** When notified about a new incoming block via websocket, the transaction appears on the Receive page but isn't shown as incoming balance throughout the rest of the application. I haven't figured how to fix that yet, or why it happens, feel free to investigate it fur me.

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -96,6 +96,7 @@
                         <div class="amount-container">
                           <div class="amount-integer">{{ (wallet.selectedAccount !== null ? wallet.selectedAccount.balance : wallet.balance) | rai: 'mnano,true' | amountsplit: 0 }}</div>
                           <div class="amount-fractional">{{ (wallet.selectedAccount !== null ? wallet.selectedAccount.balance : wallet.balance) | rai: 'mnano,true' | amountsplit: 1 }}</div>
+                          <div class="amount-updating" *ngIf="wallet.updatingBalance"><div uk-spinner="ratio: 0.5;"></div></div>
                         </div>
                       </div>
                       <div class="balance-container converted" *ngIf="settings.settings.displayCurrency">
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -89,7 +89,7 @@ export class AppComponent implements OnInit {
       this.wallet.selectedAccount = currentUpdatedAccount;
     }
 
-    await this.walletService.reloadBalances(true);
+    await this.walletService.reloadBalances();
 
     // Workaround fix for github pages when Nault is refreshed (or externally linked) and there is a subpath for example to the send screen.
     // This data is saved from the 404.html page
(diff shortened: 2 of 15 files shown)


keerifox:
added a notice when viewing a not yet confirmed block:

image

EvoSteven:
Did some light testing on this, seems to work well. Haven't found any issues.

The balance does get reset to 0 when the node server you select doesn't work. Node seems to return 200 and `{ "error": "Internal server error. Could not reach node." }` so yea.
In that case, it should probably not show any balance at all (like offline mode?), or an error message in that area, but that's an issue that might extend past the scope of this PR. Definitely worth fixing though.

Joohansson (approved):

Joohansson:
> Did some light testing on this, seems to work well. Haven't found any issues.
> 
> The balance does get reset to 0 when the node server you select doesn't work. Node seems to return 200 and `{ "error": "Internal server error. Could not reach node." }` so yea.
> In that case, it should probably not show any balance at all (like offline mode?), or an error message in that area, but that's an issue that might extend past the scope of this PR. Definitely worth fixing though.

Good suggestion, we will look at that separately. Thanks for testing, I didn't find any issues either

PR #359, opened 2021-03-31
Transaction Details and Send Confirmation redesign

keerifox:
old:

image

new:

image

\-

old:

image

new:

image


Joohansson (approved):
Looks nice and the code ok according to my eyes!

PR #363, opened 2021-04-18
Multisig and keypair generator with offline support

Joohansson:
Utilizing the existing remote signing method for creating and signing blocks with offline support gives the same ultra-secure features to the new multisig signing method.

**New Multi Signature page under advanced tools**

- Create multisig address
- Create blocks (SEND, RECEIVE, CHANGE) to be used with multisig (utilizing old remote signing method from account details page)
- Entry point for multisig participants
- Example of an M of N sharing scheme for 3 participants is given

Interface

image

Creating multisig address

8uF0vUygrW

Create a receive block

BQTe5L795o

**Keypair generator page under advanced tools**

- Generates seed+mnemonic+priv key+address
- Can be used for remote signing or new multisig addresses
- Can be used for creating offline paper wallet or other dev experiments

IIyrUognOv

**New multisig signing method in the remote signing screen**

- Any number of participants: Can be on each side of the planet with no sensitive data shared
- Multi-tab mode: To make the data sharing process automatic if using local browser tabs for all private keys used on the same machine
- Integrated with current remote signing workflow allowing the block to be published as before
- Fully compatible with Keytools multisig: Can use one Nault and one Keytools since the data format is the same
- Semi compatible with the [original musig](https://github.com/PlasmaPower/musig-nano): Can't use it at the same time but the multisig account is created the same way and funds can be recovered if Nault and Keytools would stop existing
- Full offline support meaning no private keys ever need to be exposed

Multisig manual workflow

Cutc9JlpTe

Automatic workflow using multi-tab

OE1XLnC9pH

TODO:

- [x] Touch-up after feedback, if any
- [x] Validate all functions in desktop app
- [x] Write Nault documentation with full multisig guide (update link in Nault)
- [x] Create demo video and publish to youtube


Joohansson:
Currently there is a loading problem of the musig library in the desktop app. Will have to find a solution for that before this can be merged
Edit: Solved via environment var

keerifox (approved):
quite large commit so can't vouch for it in its entirety, but it's mostly looking good on the surface (code and UX-wise)

i've suggested some wording improvements elsewhere and most of them were applied

PR #366, opened 2021-04-22
Correct backup of a private-key-wallet

### LINKED ISSUE #305: Invalid backup provided for single account wallets
Settings -> Manage Wallet -> Backup Wallet section

private key is called a seed

mnemonic is invalid, you get non-matching accounts whether you import it as nano or bip39 mnemonic

Joohansson:
**Problems before this update**
When importing a wallet as a private key or expanded private key
- Backup possible of an invalid mnemonic (based on inexisting seed)
- Bad name of backup (seed instead of private key)
- Exported wallet file is treated as a seed when imported again, resulting in wrong account. Nault can't know which wallet type.

**Fixes with this PR**
- Removes invalid mnemonic
- Correct naming of backup (private key)
- Added additional wallet types for backup file (privateKey/expandedKey)
- Handle import of new backup file format. If priv key or expanded key, treat it as such

**Notes**
- Previously exported backup files (with a seed wallet) are unaffected
- Previously exported backup files containing private key will continue to be imported as a seed and wrong account (because there is no way to differentiate them). If the user needs to recover this, they can do the import, then export the seed (which is a private key) and import that again as a private key. Now it will be treated correctly and they can save a new backup file.
- If the user has a mnemonic that was created from a private key, they can convert it using tools.nanos.cc to a seed and treat the seed as a private key to be imported into Nault
- Any backup file after this fix exported as a private key will be correctly imported

Fixes #305 

Before:
image

After:
image

### DIFF
--- a/src/app/components/configure-wallet/configure-wallet.component.ts
+++ b/src/app/components/configure-wallet/configure-wallet.component.ts
@@ -376,7 +376,8 @@ export class ConfigureWalletComponent implements OnInit {
       const fileData = event.target['result'] as string;
       try {
         const importData = JSON.parse(fileData);
-        if (!importData.seed || (!importData.hasOwnProperty('accountsIndex') && !importData.hasOwnProperty('indexes'))) {
+        if ((!importData.seed && !importData.privateKey && !importData.expandedKey) ||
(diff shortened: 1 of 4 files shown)


keerifox (approved):
code looks good i believe

PR #367, opened 2021-04-22
Progressive web app

### LINKED ISSUE #334: Progressive Web App (PWA)
Perhaps as a progressive web app?

I want to be able to turn my old Android phone into an offline signing device. With all the QR code options for remote signing, a slick experience like the Ellipal hardware wallet would be possible.

In theory we can use an Android device via a web browser, but if the browser crashes or the phone switches off and you have to switch the internet back on to reload nault, the whole purpose of remote signing is defeated.

A simple Android app designed only to be used for offline signing, intended as a companion to Nault would also work. It could even scream at you if it ever detects an internet connection :D

aleksre:
The PR turns Nault into a PWA with a service worker, installable and working offline on Android, iOS and other platforms.

Installability is promoted with a button at the bottom of the menu. Clicking it will install the app to the device. It is only shown if supported and not already installed. Unfortunately, iPhones and iPads don't support this behavior so it instead shows installation instructions (it must be done by the user in the Safari UI).

Since Nault is already available for desktop, we probably should not promote installing the PWA version on these platforms, so a whitelist is used to promote installability only on certain platforms. Currently set to Android, iOS, iPadOS and ChromeOS. If Teslas or SmartTVs start supporting PWAs in the future we can extend this.

The UX of the install widget we can discuss. It is kind of annoying currently that the address bar is static until you try to scroll to the very bottom. Maybe it would be better above/below the search box, or somewhere else entirely.

**Note:** The service worker is disabled in dev mode, so to test from localhost you must build and serve the app manually. 
A deployed version of this branch is available [HERE](https://nault-git-progressive-web-app-aleksre.vercel.app/). You can test this on your mobile devices or chromebooks.

image  image

Closes #334


aleksre:
I think I've done all the changes I wanted at this point. Moved the install widget up to above the search bar. Also added request caching. Unfortunately only GET requests can be cached by the service-worker, so most of the requests, which are POST and web socket requests, won't work offline. One option would be to implement our own request caching system (with IndexedDB for example), but I think it's fine as is with "Offline mode" as before.

image

Joohansson (approved):
Looks good to me, nice work!

keerifox (approved):
seems good

BitDesert (approved):
Works great! Thanks!

PR #386, opened 2021-05-19
Ledger iOS support - Testers wanted

aleksre:
Status: **UNTESTED** - Needs verification before merging.

This PR is an attempt at adding Ledger Nano S/X support for iOS and iPadOS version 13.3+ **via USB**.

A deployed version of the branch is available for testing here: https://nault-git-ledger-safari-support-aleksre.vercel.app/

To get this working, the `u2f-api` package, which itself is a dependency of the deprecated `@ledgerhq/hw-transport-u2f` library, had to be force resolved to the latest version 1.2.1 with Safari support. What's more, this release seems to be bugged and fails the Safari user agent check, so I created a patch for it. 
In other words, quite a few hoops we had to jump through to get this (potentially) working.

In theory this should also work for MacOS Safari, but on my M1 MacBook Pro the connect button is still disabled 🤷


nvitone23:
I just tried it with an adapter and it didn’t work. Will see if I can get a direct USB C to lightning cable.

aleksre:
Thanks for testing. Was an error message displayed? Feel free to post a screenshot.

nvitone23:
5ABBEE4D-5128-4C04-B13D-D12024DF3D2E

aleksre:
Alright, so same as if nothing is connected at all. Thanks again!

PR #399, opened 2021-06-04
Allow importing UTF-16 from address book file

Joohansson:
Forgot to allow also the file to contain non-ascii


keerifox:
is this backward compatible with old file exports?

Joohansson:
> is this backward compatible with old file exports?

Yes, the file is not in binary format, only the QR and URL. The backward compatibility is in the import-address-book.component. This step just put it in the same format as the QR before navigating to the importer page. Without the binary, the btoa fails.

PR #403, opened 2021-06-08
UI improvements for small screens

igorcmelo:
It is easier to explain with a video.
363px screen, before and after:

https://user-images.githubusercontent.com/59902451/121219061-b2107680-c859-11eb-81d3-52e4a2a9e9c2.mp4

### DIFF
--- a/src/app/components/send/send.component.css
+++ b/src/app/components/send/send.component.css
@@ -57,6 +57,7 @@
   padding-bottom: 15px;
 }
 .send-types-frame {
+  width: 100%;
   display: inline-block;
 }
 .send-types-tab {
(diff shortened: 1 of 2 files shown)


keerifox:
rather hacky solution and it introduces several new issues

i'll see what can be done to improve the tiny-width mobile layout when i have some free time

igorcmelo:
@keerifox 
What issues?
Maybe I can find out a better solution.

igorcmelo:
Also, change font sizes and paddings for smaller screens is something I see very often in web projects.

keerifox:
changes to the send page seem okay, perhaps use `uk-width-1-1` class rather than setting the width to 100%

accounts list i'll work on later as it needs a different layout

Joohansson:
What's the status of this PR? @igorcmelo do you plan to make the change of class suggested by keeri?

keerifox:
i would ask to not apply the changes to `nault-theme.less` except the last few lines concerning `send-types-tab`

PR #404, opened 2021-06-09
PWA instructions in the README

Joohansson:
Instructions were missing for PWA and users might not know about the offline possibility.


aleksre (approved):
Nice

PR #405, opened 2021-06-16
Fixing bugs in address book

### LINKED ISSUE #314: Outdated QR/URL when exporting address book
do this after #309 is merged

If you generate the address book export QR code/URL and then change something in address book, then navigate to export page, the old QR/URL will remain visible

1. The export section should be minimized when you switch panel to address book and the QR/URL variables reset, so the user needs to press "EXPORT ADDRESS BOOK" button again

2. the footer with "EXPORT ADDRESS BOOK" needs to be hidden when QR is already visible

3. IMPORT / EXPORT link must be hidden when adding a new contact

Joohansson:
Fixes #314


keerifox (approved):
code looks good

PR #406, opened 2021-06-16
Auto-remove pending while receiving

### LINKED ISSUE #402: Incoming transaction not disappearing from the Receive page
1. Have automatic receives on
2. Have the Receive page open
3. Send some nano to the account
4. Transaction appears, is automatically received, but not being removed from the list

Example: https://youtu.be/lF4u5pr1RsY?t=32

Joohansson:
This removes pending from the list while auto receiving them. This is done by calling the subscription function for pending blocks from the reloadBalances function (via clearPendingBlocks). It's not a perfect solution since the reload function first clears the pending list and then adds back any pending that exist. This means the list will flash for some milliseconds before the pending are added back in. Would be better if each item was more nicely removed but it's complicated. The advantage of doing it this way is it will also work if some other wallet is receiving at the same time. The pending will always be correct.

removePendingBlock() is not used anywhere in the wallet but added it there as well in case it will be used in the future.

Fixes #402

### DIFF
--- a/src/app/components/receive/receive.component.html
+++ b/src/app/components/receive/receive.component.html
@@ -90,7 +90,7 @@
           </tr>
           </thead>
           <tbody>
-          <tr *ngFor="let pending of pendingBlocksForSelectedAccount" [class]="{ 'uk-text-muted': true, 'uk-hidden': pending.received }">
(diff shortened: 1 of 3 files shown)


keerifox (approved):
code looks good

PR #416, opened 2021-07-12
Stop fetching active_difficulty from the network

### LINKED ISSUE #407: Deprecate active_difficulty
calls to `active_difficulty` RPC are no longer required since the difficulty is fixed at 1x

Joohansson:
Deprecated function since node v22

Fixes #407

### DIFF
--- a/src/app/components/configure-app/configure-app.component.ts
+++ b/src/app/components/configure-app/configure-app.component.ts
@@ -127,7 +127,6 @@ export class ConfigureAppComponent implements OnInit {
 
   multiplierOptions = [
     { name: 'Default (1x or 1/64x)', value: 1 },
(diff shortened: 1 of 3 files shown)


keerifox:
does this handle users having "Automatic (max 8x)" already set as their option?

Joohansson:
> does this handle users having "Automatic (max 8x)" already set as their option?

Yes, if the user has that option saved in the settings, it will be detected as unmatched and defaults to the first option (Default). https://github.com/Nault/Nault/blob/38d87d13a01f4953ad7f07fd5ed6ea3257051bd5/src/app/components/configure-app/configure-app.component.ts#L270

keerifox (approved):
recall something like that but didn't have time to check myself

code seems to be fine

PR #417, opened 2021-07-16
Mobile-friendly account view (new homepage)

keerifox:
This PR adds a mobile-friendly version of the account page, which can be accessed...
\- by clicking on the account from the Accounts page – this will now automatically select the account in sidebar (mobile only)
\- by clicking on the account name from the active account card in sidebar
\- by starting Nault (if your wallet is configured and you have an account selected in the sidebar)
\- by using the "Back" button to navigate to the previously visited pages

Note: this page variant is only for accounts belonging to the configured wallet

To switch from compact view to full account details, simply press on ~~the name of the account~~ the 3-dots icon and choose the "View account details" option

image

Change to desktop homepage: Instead of "Accounts" page, the "Account Details" page will open, if the wallet is configured and an account is selected in sidebar

Notifications will appear slightly higher on small viewports, to prevent overlap with the buttons

This PR also fixes "Welcome" and "Accounts" pages being left in the browser history when the app redirects away from them to the actual page that was requested


Joohansson (approved):
Tested on desktop with a narrow view to simulate mobile. Looks good!

PR #418, opened 2021-07-17
Fix not being able to scroll to the very top/bottom on mobile devices

### LINKED ISSUE #318: App layout is taller than mobile device's viewport
bug can be observed by not being able to scroll sidebar or main section to the very end (or sometimes to the very start)

uk-height-viewport seems to be one of the causes, but not the only reason the issue occurs

keerifox:
Fixes #318


Joohansson (approved):
Works on iphone, chrome and Safari

PR #432, opened 2021-07-22
Determine extra raw amount when using url param of the send page

keerifox:
Improve style of the extra raw amount
Fix fiat price not being shown when using amount url param
Fix error if passed amount that is not a number

Rename amountRaw to amountExtraRaw to reduce confusion

### DIFF
--- a/src/app/components/send/send.component.css
+++ b/src/app/components/send/send.component.css
@@ -6,6 +6,20 @@
 	max-width: 450px;
 }
 
+.amount-raw-extra {
+  margin: 12px 5px 12px 10px;
+}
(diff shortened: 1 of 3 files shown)


keerifox:
in other words, when you use the `Sends funds back` feature in the mobile interface, you will be redirected to a page like this:
https://nault.cc/send?amount=15.000000000000000000000000000001

which will prefill the following amount:

image


this PR will separate the amount that's less than `1 millionth Nano`, adding it as the "extra raw amount" which you can remove by pressing the X next to it

the nano amount field will just have "15" in it

PR #448, opened 2021-07-23
Upscale nanoidenticons when devicePixelRatio is higher than 1

### LINKED ISSUE #423: Base Nanoidenticons size on device pixel ratio
currently Nanoidenticons appear lowres on mobile devices, perhaps multiply their size by device pixel ratio and `Math.ceil` it to the nearest integer

keerifox:
Fixes #423

to test this:

1. open a nault page with nanoidenticons (e.g. transactions list)
2. zoom in to ~170%
3. observe that the identicons are blurry

image

4. refresh, causing new identicons to generate
5. observe that the identicons are crisp

image


Joohansson:
Works fine!

PR #455, opened 2021-07-25
Use the new style for confirming remote signing

### LINKED ISSUE #427: Process Nano Block page should use new layout
Current layout:

image

Expected layout:

image

Joohansson:
Fixes #427 

* Fiat amount hidden if signing on an offline device
* Another detail that differs from the original SEND screen is the "N/A balance" of the sender when receiving funds. That's how the nano protocol is designed, the info is not part of the block. Plus it's forced to N/A also for the recipient balance to avoid confusion if the recipient balance changes while doing the signing of SEND blocks (which it's allowed to do)

Example screens:
image

image

image

### DIFF
--- a/src/app/components/sign/sign.component.css
+++ b/src/app/components/sign/sign.component.css
@@ -101,3 +101,49 @@
 .add-button span {
   vertical-align: text-bottom;
 }
+
+.block-amount-primary {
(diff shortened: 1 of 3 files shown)


keerifox (approved):
didn't test locally but the code/screenshots look good, just the linebreak needs fixing

image

PR #460, opened 2021-07-29
Translation account details

BitDesert:


src/app/components/account-details/account-details.component.html
```diff
@@ -385,7 +385,7 @@ <h3 class="uk-margin-remove-bottom">Transactions</h3>
             </ng-template>
           </ng-template>
           <td class="amount-column uk-text-middle uk-text-muted" [title]="('Incoming Transaction') + ( (pending.amountRaw && (pending.amountRaw > 0) ) ? ( ', +' + ( pending.amountRaw.toString(10) ) + ' raw' ) : '' )">
-            <span class="uk-text-small">Ready to receive</span><br>
+            <span class="uk-text-small">{{ 'account-details.ready-to-receive' | transloco }}</span><br>
```
keerifox: could use `'general.actions.ready-to-receive'` here

src/app/components/account-details/account-details.component.html
```diff
@@ -724,7 +724,7 @@ <h3 class="menu-header">{{
           : (
               walletAccount
             ? (('general.account' | transloco) + '#' + walletAccount.index)
```
keerifox: missing a space before `#` here.

could use a utility function for getting account label perhaps? or like, a pipe, i don't know

i have one on the receive component (but it's needed across at least 4 components so maybe it could be in a single file)
https://github.com/Nault/Nault/blob/bbb79ca600d67b774332c6cd43b7d534e89e214e/src/app/components/receive/receive.component.ts#L172-L180

(which is also missing a space now)

src/app/components/account-details/account-details.component.html
BitDesert: Fixed, thanks!

src/app/components/account-details/account-details.component.html
BitDesert: Fixed! Yeah maybe a helper function available in all components would be better.

keerifox (approved):

Issue #471, opened 2021-08-29
Ensure returned responses from backend servers are valid json


keerifox:
if json response is not valid, treat it same as a server error (see catch method below)

https://github.com/Nault/Nault/blob/56e605d98936476e447005eb6ac6a1e9095fccd5/src/app/services/api.service.ts#L32-L56

keerifox:
unfortunately non-json responses still make it to `.then(res => {` as an empty object

keerifox:
seems to work

PR #473, opened 2021-08-29
Require API calls to return a JSON response

keerifox:
Potential fix for #471

Please test with a custom backend server that returns a non-JSON response (e.g. missing `}` at the end) and verify that node is considered offline as a result of the call (should autoswitch backend if set to "Random")


keerifox:
this can probably be merged

Issue #413, opened 2021-07-03
Hide Balances


NashJames:
**Is your feature request related to a problem? Please describe.**
Two purposes:
- Checking transactions without having someone view your total balances over your shoulder
- Bitvavo now requests screen recordings of you accessing your account, to verify you own the account. I would rather not share the exact balances of my accounts to their record.

**Describe the solution you'd like**
A simple toggle button, to replace any nano or fiat values with a set number of asterisks. Or something similar

**Additional context**
Ideally this would including the individual account pages and their transaction lists too. Not just the total balance in the corner

Joohansson:
There are balances in many places in the wallet such as the account detail page, account history, the main menu balance widget, send screen, receive screen etc. Should they all be hidden from an app setting or just add a toggle button to the main widget? I guess all because there is nowhere to navigate without showing balances somewhere. But maybe the main widget toggle button could also trigger a global hide. I'm not sure what's the most user friendly here.

NashJames:
I think a global hide would cover the majority of use cases, and would be a simpler implementation. The location of the toggle, main widget or settings page, is largely up to you. I suppose there is a chance of it getting frequent use, if someone wishes to only briefly view their balance for example, then hide it again.

andrew-m-higgs:
Is anyone working on this?

I have implemented a solution using configure-app.
I did a transfer out while balances were hidden and saw that the balance of the account is visible there too. Will track that one down and fix it. 
I guess the amount being transferred should not be hidden. Easier to make sure you are sending the correct amount. 
Thoughts?

geommr:
> I guess the amount being transferred should not be hidden. Easier to make sure you are sending the correct amount.

agreed. As far as I can tell, there are 4 places where the balance should/could be hidden: 
- the wallet card widget
- account details page
- list of accounts
- sending confirmation page

@andrew-m-higgs any progress on this? I could eventually work on it if not.

andrew-m-higgs:
@geommr There is a pull request in for this (https://github.com/Nault/Nault/pull/476).

Regards

PR #476, opened 2021-09-15
Hide balances #413

andrew-m-higgs:
My version of how this issue can be resolved.

Thoughts on a better way would be appreciated.


keerifox:
> Hope this does not bother anyone.

yeah umm

image


image

keerifox:
image

this option should be under Display Settings, perhaps as a dropdown selection (as it's difficult to tell it's a checkbox)

\-

image

the label on "Accounts" page needs to be filtered too

\-

and perhaps one of the most important places to filter would be this dropdown on the "Receive" page as it's used as Point-of-Sale UI

image

\-

otherwise the PR seems to be good

linting makes it painful to review the changes so i cannot vouch for everything

keerifox:
not sure if it makes sense to filter individual transaction amounts (especially for accounts not belonging to the wallet), but the mobile view is currently unaffected

image

andrew-m-higgs:
> > Hope this does not bother anyone.
> 
> yeah umm
> 
> image
> 
> image

:-0 Oops!

andrew-m-higgs:
> -
> 
> image
> 
> the label on "Accounts" page needs to be filtered too

Not quite sure where this is. Mine is always just showing the text-snippet which is 'New' with no balance. I see the text-full does include a balance which I have now hidden. Is this the place?

Also just wondering if we need the mask to be *.****** or if *.*** would be sufficient.

Thanks.

andrew-m-higgs:
> not sure if it makes sense to filter individual transaction amounts (especially for accounts not belonging to the wallet), but the mobile view is currently unaffected
> 
> image

Finally found where the layout is coming from. :-)

Weird though, when I put 'settings.settings.hideBalances ? '*' : ' in front of transaction.amount it completely destroys the layout. Any ideas?

keerifox:
> Mine is always just showing the text-snippet which is 'New' with no balance.

it depends on the width of the view

src/app/components/receive/receive.component.html
```diff
@@ -200,8 +200,8 @@ <h3 class="uk-margin-remove-bottom">Incoming Transactions</h3>
             </ng-template>
             <td class="amount-column uk-text-middle uk-text-muted" [title]="('Incoming Transaction') + ( (pending.amountRaw && (pending.amountRaw > 0) ) ? ( ', +' + ( pending.amountRaw.toString(10) ) + ' raw' ) : '' )">
               <span class="uk-text-small">Ready to receive</span><br>
-              <span class="amount-integer">{{ pending.amount | rai: 'mnano,true' | amountsplit: 0 }}</span>
-              <span class="amount-fractional">{{ pending.amount | rai: 'mnano,true' | amountsplit: 1 }}</span>
+              <span class="amount-integer">{{ settings.settings.hideBalances ? '*' : pending.amount | rai: 'mnano,true' | amountsplit: 0 }}</span>
```
keerifox: not sure how many merchants will use the manual receives option, but they kinda do need to know the amount of nano they just received, without revealing balances of their other accounts (when using nault as a point-of-sale checkout)

keerifox:
> Weird though, when I put 'settings.settings.hideBalances ? '*' : ' in front of transaction.amount it completely destroys the layout. Any ideas?

you'll have to pass the setting value similarly to how `settingIdenticonsStyle` is passed to it (there are 4 places where `app-nano-transaction-mobile` is used)

src/app/components/receive/receive.component.html
andrew-m-higgs: So perhaps we just leave the pending receives. I assume they would not have many pending?

andrew-m-higgs:
Will check it out tonight.

Thanks for the help. Much appreciated.

andrew-m-higgs:
I hope I got everything done.

Must say that getting hideBalances setting across to nano-transaction-mobile had me stumped for a while. Hopefully I have learnt something and won't struggle to much next time.

src/app/components/configure-app/configure-app.component.html
```diff
@@ -71,6 +72,23 @@ <h3 class="uk-card-title">Display Settings</h3>
             </div>
           </div>
 
+          <div class="uk-width-1-1">
+            <div class="uk-form-horizontal">
+
+              <div class="uk-margin">
+                <label class="uk-form-label" for="form-horizontal-select1">
+                  Hide Balances <span uk-icon="icon: info;" uk-tooltip title="Hide balances when viewing wallet and account details. Does not affect pending incoming amounts on the receive page."></span>
```
keerifox: the term "pending" has been replaced with "ready to receive" in UX, we shouldn't use it

keerifox:
the name `settingHideBalances` would be better than `hideBalances` (for consistency), but not critical

\-

most seems to be working fine

balances are still shown when viewing/changing reps, but that could be changed later on outside this PR if needed.

i suppose "to account" is supposed to be hidden too?

image

\-

two nano-transaction-mobile elements on the Receive page don't have the `hideBalances` prop, which doesnt hide the balances (so works as it should). perhaps value of `false` could be passed there for clarity (that balances should not be hidden, no matter the setting)

src/app/app.component.html
```diff
@@ -307,13 +330,7 @@
...
 
       </div>
-      <div
-        class="uk-width-expand content-container uk-panel"
-        [class.nav-expanded]="navExpanded"
-        [class.nav-animating]="navAnimating"
-        [style.height]="( innerHeightWithoutMobileBar + 'px' )"
-        (click)="closeNav()"
-      >
+      <div class="uk-width-expand content-container uk-panel" [class.nav-expanded]="navExpanded" [class.nav-animating]="navAnimating" [style.height]="( innerHeightWithoutMobileBar + 'px' )" (click)="closeNav()">
```
keerifox: aaaaa i gave them newlines for a reason

andrew-m-higgs: I think the latest push should fix these issues. Hopefully I got them all.

src/app/components/configure-app/configure-app.component.html
andrew-m-higgs: Done.

andrew-m-higgs:
> the name `settingHideBalances` would be better than `hideBalances` (for consistency), but not critical

Is this with reference to app-nano-transaction-mobile specifically?

keerifox:
> Is this with reference to app-nano-transaction-mobile specifically?

yup. thank you for the contributions and patience!

i think balance hiding on the account details page might vary for ready-to-receive tx between mobile view and default view (looking at code, didn't test it)

the tooltip that says "Does not affect ready to receive incoming amounts on the receive page" might need tweaking to "Does not hide amounts for individual transactions that are ready to be received" (since it's includes account details page now)

and when changing representative there is a dropdown that shows amounts, i think that's the last bit

andrew-m-higgs:
> i think balance hiding on the account details page might vary for ready-to-receive tx between mobile view and default view (looking at code, didn't test it)

Should be sorted in 1f804e1. Unless you feel that where ever a ready to receive amount is shown, it should not be hidden? Or do we just stick to the receive page?

andrew-m-higgs:
> and when changing representative there is a dropdown that shows amounts, i think that's the last bit

Is there somewhere which is not covered by d3f73b2?

andrew-m-higgs:
> yup. thank you for the contributions and patience!

No worries. I am enjoying the opportunity to give back.
The Nano community is great. The Nault community is no different. :-)

keerifox:
> 
> 
> > and when changing representative there is a dropdown that shows amounts, i think that's the last bit
> 
> Is there somewhere which is not covered by [d3f73b2](https://github.com/Nault/Nault/commit/d3f73b2444b849d681d0eaf831a310fc247078b5)?

here i think https://github.com/Nault/Nault/blob/d3f73b2444b849d681d0eaf831a310fc247078b5/src/app/components/representatives/representatives.component.html#L139

keerifox:
> Should be sorted in [1f804e1](https://github.com/Nault/Nault/commit/1f804e1c00e9538b12f9cb4cbf81990396187510). Unless you feel that where ever a ready to receive amount is shown, it should not be hidden? Or do we just stick to the receive page?

i'm not sure, i suppose it could be hidden outside the receive page.

\-

https://github.com/Nault/Nault/blob/1f804e1c00e9538b12f9cb4cbf81990396187510/src/app/components/accounts/accounts.component.html#L117-L125

missing a tab here on lines 118-121. entire code block got shifted to the right but those lines were left untouched

andrew-m-higgs:
Any thoughts on adding a means to toggle the hiding of balances?
Perhaps something like this?

image image

andrew-m-higgs:
What about balances tracked in the address book?

keerifox:
> 
> 
> Any thoughts on adding a means to toggle the hiding of balances?
> Perhaps something like this?
> 
> image image

probably not needed for now, i might get around drawing and adding an eye icon at the top of the dropdown somewhere here

image

keerifox:
> 
> 
> What about balances tracked in the address book?

not sure, don't think it's required for now

andrew-m-higgs:
> > What about balances tracked in the address book?
> 
> not sure, don't think it's required for now

I think this might be needed. What about when I have a name, say 'Bob Smith' and the balance of their account?

andrew-m-higgs:
> 
> probably not needed for now, i might get around drawing and adding an eye icon at the top of the dropdown somewhere here
> 
> image

Not quite where you suggested but what about: 
image image
The SVG iteslf can of course be improved upon. I am no artist.

keerifox:
> 
> 
> > probably not needed for now, i might get around drawing and adding an eye icon at the top of the dropdown somewhere here
> > image
> 
> Not quite where you suggested but what about:
> image image
> The SVG iteslf can of course be improved upon. I am no artist.

i wouldn't add anything there because it's veeery crowded already

andrew-m-higgs:
> i wouldn't add anything there because it's veeery crowded already

Okay. Not sure I agree with that. :-)

Where would you suggest? I think it should be a one click procedure. Unless, of course, it is on mobile where you first need to get to the menu.

Joohansson:
@andrew-m-higgs There seems to be a merge conflict with the master branch. Could you please try and resolve those and I'm looking to merge this whole PR. I don't feel confident resolving those myself. You can probably make a rebase from master or merge master into your branch.

andrew-m-higgs:
One small conflict left. Not even sure why it is there. Looks pretty straight forward. Any help resolving this last one would be great.

I am no git expert. If this is not right then I would be more comfortable making a new pull request.

Thanks.

BitDesert:
@andrew-m-higgs I don't know exactly why this conflict happened but it's resolved now.

andrew-m-higgs:
> @andrew-m-higgs I don't know exactly why this conflict happened but it's resolved now.

Thank you kindly.

Joohansson:
The PR looks good. There are two more places that display balances.

1. In the address book if tracking watch only accounts. Both in the table and under "Show total tracked balance". I think we can skip this since it could be either local or remote accounts. It's up to the user.
2. Under advanced tools -> wallet sweeper -> local destination dropdown. Maybe we want this too while at it @andrew-m-higgs ?

andrew-m-higgs:
> The PR looks good. There are two more places that display balances.
> 
>     1. In the address book if tracking watch only accounts. Both in the table and under "Show total tracked balance". I think we can skip this since it could be either local or remote accounts. It's up to the user.
> 
>     2. Under advanced tools -> wallet sweeper -> local destination dropdown. Maybe we want this too while at it @andrew-m-higgs ?

Yes. I did mention the address book one. Maybe we should just hide both instances. Should I just hide both?

Joohansson:
> > The PR looks good. There are two more places that display balances.
> > ```
> > 1. In the address book if tracking watch only accounts. Both in the table and under "Show total tracked balance". I think we can skip this since it could be either local or remote accounts. It's up to the user.
> > 
> > 2. Under advanced tools -> wallet sweeper -> local destination dropdown. Maybe we want this too while at it @andrew-m-higgs ?
> > ```
> 
> Yes. I did mention the address book one. Maybe we should just hide both instances. Should I just hide both?

I'm fine with either. Go ahead!

andrew-m-higgs:
Done.

BitDesert:
How about using a pipe? Then the code would be better readable IMO:

`{{ wallet.balance | rai: 'mnano,true' | hidebalance: settings.settings.hideBalances | amountsplit: 0 }}`

If we need a different count of stars for the fraction it would look like this:

`{{ wallet.balanceFiat | fiat: settings.settings.displayCurrency | hidebalance: settings.settings.hideBalances:2 }}`

What do you think? Here is a quick test: https://github.com/Nault/Nault/commit/9688239ff60e05f506cbbdfd54cf6a1220cdbb10

keerifox:
mentioning @andrew-m-higgs for the comment above

PR #480, opened 2021-09-26
Added french localization

le-jun:
French localization from the content of en.json


BitDesert:
Thank you for your contribution! We're currently in the process of converting all subpages into the new format so they can get translated. So more strings are coming soon.

PR #481, opened 2021-09-27
Welcome translation

BitDesert:


keerifox:
image

this PR removes all of those necessary in-text links

BitDesert:
Some in-text links got moved to a "Learn more" link at the bottom. 
This change is necessary because if you want to use in-translation links you would have to allow dynamic html in the template which is a huge security issue.
That's why all in-line links need to be rewritten.

keerifox:
if it's possible to retain the current look/functionality ( by doing similar as described here https://github.com/Nault/Nault/pull/482#issuecomment-929019622 ), that'd be very appreciated! a lot of thought went into it to try to keep it accessible and easy to understand

BitDesert:
@keerifox fixed :)

PR #482, opened 2021-09-27
Accounts translation

BitDesert:


keerifox:
this PR diminishes user experience by removing links from tips/placeholders

image

BitDesert:
Same reason as in https://github.com/Nault/Nault/pull/481#issuecomment-928899404

But if we really need it you could split the string in 3 parts which could be translated individually (each language could have a different order).

keerifox:
ideally it would be `that's a {{link}} mid text` where `{{link}}` references a `<a href="linkhere">{{link_text}}</a>` (only the plain text exposed to transloco, no unsafe html would have to be used)

```
"link_sentence": "that's a {{link}} mid text",
"link_text": "clicky"
```

but i'm not sure if this functionality is available in transloco

\-

splitting into parts could work as a workaround. i don't like it very much but it's better than reducing UX (in both PRs) for the sake of translation

```
"link_sentence": {
    "1": "that's a",
    "2-link-clicky": "clicky",
    "3": "mid text"
}

"link_sentence": {
    "1": "mid text, that's a",
    "2-link-clicky": "clicky",
    "3": ""
}

"link_sentence": {
    "1": "",
    "2-link-clicky": "clicky",
    "3": "that is, mid text"
}
```

BitDesert:
@keerifox the links are now clickable again.

keerifox (approved):
thank you, the rest also seems good i believe

PR #483, opened 2021-09-28
Add transloco test to make sure that you cannot use the language word…

### LINKED ISSUE #463: Translations and address book labels
**Describe the bug**
You can rename an account to "Konto #5"

**Expected behavior**
You should not be able to rename an account to "Konto #5" (regardless of which language is used)

**Additional context**
https://github.com/Nault/Nault/blob/bbb79ca600d67b774332c6cd43b7d534e89e214e/src/app/components/account-details/account-details.component.ts#L554
https://github.com/Nault/Nault/blob/bbb79ca600d67b774332c6cd43b7d534e89e214e/src/app/components/address-book/address-book.component.ts#L277

andrew-m-higgs:
… set in general.account followed by space#number as the name used in Address Book.

Fixes #463 assuming that just the current chosen language is restricted.

A potential loophole is that if I set to 'Account #1' while in German, it will stay when setting language to English.

### DIFF
--- a/src/app/components/account-details/account-details.component.ts
+++ b/src/app/components/account-details/account-details.component.ts
@@ -551,7 +551,7 @@ export class AccountDetailsComponent implements OnInit, OnDestroy {
       return;
     }
 
-    const regexp = new RegExp('^Account|' + this.translocoService.translate('general.account') + ' #\\d+$', 'g');
+    const regexp = new RegExp('^(Account)|(' + this.translocoService.translate('general.account') + ') #\\d+$', 'g');
     if ( regexp.test(this.addressBookModel) === true ) {
       return this.notifications.sendError(`This name is reserved for wallet accounts without a label`);
     }
--- a/src/app/components/account-details/account-details.component.ts
+++ b/src/app/components/account-details/account-details.component.ts
@@ -551,7 +551,8 @@ export class AccountDetailsComponent implements OnInit, OnDestroy {
       return;
     }
 
-    if ( /^Account #\d+$/g.test(this.addressBookModel) === true ) {
+    let regexp = new RegExp('^' + this.translocoService.translate('general.account') + ' #\\d+$', 'g');
+    if ( regexp.test(this.addressBookModel) === true ) {
       return this.notifications.sendError(`This name is reserved for wallet accounts without a label`);
     }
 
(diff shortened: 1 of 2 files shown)


keerifox:
hmm perhaps it could be (Account|\<word Account in current locale\>) so that even if there is a loophole it doesn't apply to english

i don't know if it's reasonable/feasible to loop through each language and check absolutely all of them

andrew-m-higgs:
> hmm perhaps it could be (Account|<word Account in current locale>) so that even if there is a loophole it doesn't apply to english
> 
> i don't know if it's reasonable/feasible to loop through each language and check absolutely all of them

I like this suggestion.
Yes, I am sure it is not worth trying to loop through all detected languages.

keerifox (approved):
seems ok

could have parenthesis in the regexps to make it more clear the OR condition applies to the whole word and not just the letter `t` of Account and the first letter of translated word

another reason to have parenthesis is if a translation of word Account ends up having a space in some language (let's say a fictional language where it's written as "akke counte", the regex would expect Account OR akke, so english version would not match unless you type "Account counte #2"

andrew-m-higgs:
> could have parenthesis in the regexps to make it more clear the OR condition applies to the whole word and not just the letter `t` of Account and the first letter of translated word

Good point.

Sommai242323 (approved):

BitDesert:
I'm not sure if we should build the regex dynamically because there are some special characters in regex which could lead to some weird behaviour. But this shouldn't be a security issue.

Do we even need the check? Why shouldn't we allow any label?

andrew-m-higgs:
> I'm not sure if we should build the regex dynamically because there are some special characters in regex which could lead to some weird behaviour. But this shouldn't be a security issue.
> 
> Do we even need the check? Why shouldn't we allow any label?

I personally don't mind either way. I suppose the only thing you really should be sure of is that 2 accounts don't have the same name.

What would happen though, if I have 1 account (default label would be 'Account #1') I then decide to name it 'Account #2' (perhaps because it is my second wallet) and then want to create another account (using index 2)?

BitDesert:
> What would happen though, if I have 1 account (default label would be 'Account #1') I then decide to name it 'Account #2' (perhaps because it is my second wallet) and then want to create another account (using index 2)?

Yeah could be confusing, but maybe it's your own fault then. But maybe it's better to have that check for now.

BitDesert:
Thanks for the PR @andrew-m-higgs !

andrew-m-higgs:
Absolute pleasure. Thank you to both @keerifox  and @BitDesert for all the help and assistance along the way.

PR #484, opened 2021-10-02
Translation - Configure app

BitDesert:


keerifox:
i feel it might be unclear to users that "Save" button only saves that specific category. the notification message doesn't clear that misconception either

BitDesert:
> 
> 
> i feel it might be unclear to users that "Save" button only saves that specific category. the notification message doesn't clear that misconception either

Should both messages (button + notification) include the category? Or just the button?

keerifox:
> 
> 
> > i feel it might be unclear to users that "Save" button only saves that specific category. the notification message doesn't clear that misconception either
> 
> Should both messages (button + notification) include the category? Or just the button?

probably both for clarity, otherwise the Save buttons would need to save all of the categories and/or have a state that's shown when there are unsaved changes (e.g. button primary "Save" and button disabled "Saved")

PR #490, opened 2021-10-14
Change 'Unknown' to 'Unknown Rep'

### LINKED ISSUE #487: Unknown Rep label translation
image

this should say "Unknown Rep" instead of "Unknown"

in some languages "Unknown" may need to be spelled differently based on the adjacent word (he/she/it), which makes it difficult to fit in that sentence

even in English "Unknown represents you..." sounds off compared to what it used to be "Unknown Rep represents you..."

andrew-m-higgs:
Fixes #487 
Add transloco field for 'Unknown Rep' under 'change-rep-widget.unknown-rep'. 
Use this field in the rep details for unknown representative.

### DIFF
--- a/src/app/components/change-rep-widget/change-rep-widget.component.html
+++ b/src/app/components/change-rep-widget/change-rep-widget.component.html
@@ -12,7 +12,7 @@
       : ('general.representative' | transloco)
     }}</h2>
     <div class="representative-name-row">
-      <a (click)="showRepSelectionForSpecificRep(rep)" class="name">{{ rep.label || ('general.unknown' | transloco) }}</a>
+      <a (click)="showRepSelectionForSpecificRep(rep)" class="name">{{ rep.label || ('change-rep-widget.unknown-rep' | transloco) }}</a>
       <div class="weight-total" *ngIf="!rep.percent.isZero()">{{ rep.percent.toFixed(2) }}%</div>
     </div>
     <ng-container [ngSwitch]="true">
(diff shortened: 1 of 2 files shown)


keerifox:
https://github.com/Nault/Nault/blob/565f15dfe441c0eff1b991dbf20a1f5b2ba26e7d/src/app/components/change-rep-widget/change-rep-widget.component.html#L15

needs to be replaced here too i believe, otherwise there's an inconsistency, and the issue with translation forms is still present in 1/2 places

andrew-m-higgs:
> https://github.com/Nault/Nault/blob/565f15dfe441c0eff1b991dbf20a1f5b2ba26e7d/src/app/components/change-rep-widget/change-rep-widget.component.html#L15
> 
> needs to be replaced here too i believe, otherwise there's an inconsistency, and the issue with translation forms is still present in 1/2 places

Fair enough. I actually left it because I thought it could stay as is. But this probably is better.

BitDesert:
"Rep" is a abbreviation that only works well in english. Other languages would have to shorten their term as well, is this desired?

andrew-m-higgs:
> "Rep" is a abbreviation that only works well in english. Other languages would have to shorten their term as well, is this desired?

This can be set to whatever the other language requires. They just change it in their language file. In English it is set to two words another language might only need one word.

Another option could be to change to a quoted rep name. 

"'Unknown' represents you..." or "'NanoLooker' represents you..."

then we can just leave it as is.

keerifox (approved):

BitDesert:
> 
> 
> > "Rep" is a abbreviation that only works well in english. Other languages would have to shorten their term as well, is this desired?
> 
> This can be set to whatever the other language requires. They just change it in their language file. In English it is set to two words another language might only need one word.
> 
> Another option could be to change to a quoted rep name.
> 
> "'Unknown' represents you..." or "'NanoLooker' represents you..."
> 
> then we can just leave it as is.

Good point :thumbsup:

PR #492, opened 2021-10-17
Fix error message when max trackable accounts is reached.

### LINKED ISSUE #491: Max trackable error message notreporting the number
**Describe the bug**
The error message which pops up after reaching the maximum trackable accounts says 'You can only track up to accounts'

**To Reproduce**
Keep adding accounts until maximum is reached and the error message pops up.

**Expected behavior**
The error message should read 'You can only track up to 20 accounts'

andrew-m-higgs:
Fix error message which pops up when reaching the maximum trackable accounts.
Fixes #491.


BitDesert:
Thanks @andrew-m-higgs !

PR #504, opened 2021-11-29
Translation - Account details + Receive

BitDesert:


src/app/components/receive/receive.component.html
```diff
@@ -260,21 +260,21 @@ <h3 class="uk-margin-remove-bottom">Incoming Transactions</h3>
             *ngIf="!mobileTransactionData.received"
           ></span>
           <span *ngIf="!mobileTransactionData.received && !mobileTransactionData.loading">Receive funds</span>
```
keerifox, line 262: code changes look good i believe, only missing a translation for this string

PR #507, opened 2021-12-23
Prompt wallet password instead of stating that "Wallet must be unlocked"

### LINKED ISSUE #260: Prompt wallet password instead of stating that "Wallet should be unlocked"
nice-to-have UX feature that can be seen in other wallets

keerifox:
Fixes #260

Replace `walletService.walletIsLocked()` with `walletService.isLocked()` as they seem to be serving the same purpose
Clear password variable/input when hiding wallet unlock modal (previously only cleared upon modal show / wallet unlock)
Wallet unlock modal is now bottom-aligned on small viewports
All other bottom-aligned modals are now only bottom-aligned on small viewports


BitDesert:
Great addition, thanks @keerifox !

PR #518, opened 2022-02-06
Improve Ledger UX

### LINKED ISSUE #516: Remove warning when replacing a Ledger wallet
Whenever the currently configured wallet is a Ledger device and the user tries to configure a new wallet, the following warning is displayed:

image

This is displayed regardless of what the new wallet is. It could be the same Ledger device, a different Ledger device, a new hot wallet, or an existing hot wallet.

This big red warning sends a strong message to the user that they should think closely about what they are about to do. A warning like this should not be overused as it might desensitize the user to warnings. Such a warning is justified when the user tries to replace a seed/private key based wallet of which they may not have a backup, but IMO *not* justified when replacing a Ledger based wallet for the following reasons:

- Since the current wallet is a Ledger device, we know that the user has both access to their Ledger and the PIN-code. This means replacing the wallet is a completely harmless operation that they shouldn't be "warned" against. They can simply reconnect it at any time. Even if the user has since lost access to the device or the PIN, stopping them from configuring a different wallet, will not help at all. The Ledger wallet cannot be used until it is reconnected.
- Informing the user about the importance of their Ledger seed/mnemonic probably creates more confusion than it clarifies. It is completely impossible to complete the Ledger setup process without writing down your mnemonic, and ensuring that the user understands its importance is fully on Ledger, not Nault. Without a seed backup, all crypto on the Ledger is at risk of being lost, not just the Nano wallet. Some users might think there is something specific about Nault they need to write down and make a copy of, while in reality they already did everything right. Even if they somehow destroyed their backup, stopping them from configuring a new wallet will not help.
- In many cases the user might be trying to connect the very same Ledger device that Nault remembered from last time. Seeing this warning then, is particularly odd.

I therefore suggest we remove this warning.

aleksre:
This PR improves many aspects of the Ledger integration. The most notable of which are:

- Bluetooth support for Nault Desktop
- Windows no longer prefers WebHID over WebUSB. The inconsistencies with WebUSB seems to be fixed and it now works better than WebHID, which seems to have trouble reconnecting after the device is locked.
- The "Refresh Ledger status" button actually works now. Before it would just display a success notification if `ledger.status` was `'ready'`, which was the condition for displaying the button in the first place. Now it will ping the ledger and update its status.
- Connection timeout set to 30 seconds for all transports (including desktop)
- Poll interval reduced to 5 seconds for all transports except U2F. This means that when disconnecting the device, Nault will update its status to "Not connected" within a few seconds. U2F increased from 15 to 30 seconds, which means the annoying popup is shown less frequently.
- Fixed a bug that prevented the "Ledger Locked" status from showing up (it would go straight to "Not connected")
- ~~Replaced~~ Changed the warning modal when replacing a Ledger-based wallet. Closes #516.
- Changed the text of several Ledger related messages to be more clear.
- Updated Ledger libraries.

**Note:** The Bluetooth support for desktop is quite particular about which BT modules it supports. Because of this I don't think we should necessarily recommend people who rely on BT support to use the desktop release. On my Windows computer, BT does not work in the desktop version, even though it works great in Chrome and Edge. On my Mac however, it works also in the desktop version. Might be a good idea to write something about this in the troubleshooting guide. There is a message displayed if you have an incompatible BT module in the desktop version, but we wont know if it works until the user tries to connect unfortunately.

**Also note:** I have **not** been able to reproduce the bug where someone got a success message on the ledger connection screen, but a not connected message in the side-bar. I have tried everything I can think of, but we need more info about the exact steps to reproduce this if we want to fix it.

Branch deployed to: https://nault-git-ledger-ux-aleksre.vercel.app


keerifox:
most of what i've been able to test/review seems to be fine

my only complaint is that with the modal confirmation removed it's now too easy to miss the "you are about to stop using ledger" notification and press "new wallet", thinking you're still using ledger. i'd keep that confirmation dialogue, even if it's only in info color

that's what i meant here https://github.com/Nault/Nault/issues/516#issuecomment-1014905230 since the context of the post was the confirmation modal

Joohansson:
Re-trigger checks

aleksre:
Sorry for the delay. PR should be good now. The modal when replacing a Ledger wallet is back in info color.

PR #524, opened 2022-03-08
Add hash/navigation to successor block on the Transaction Details page

keerifox:
This allows to navigate to the next block (e.g. from height 108 to 109)

Requires V23.0+ backend server: https://docs.nano.org/releases/release-v23-0/#rpc-updates

image

`Height` field has been moved above `Link` field to avoid accidental clicks on `Link` field when trying to navigate to successor of the last block in chain (which it doesn't have)


BitDesert:
Thanks @keerifox !

PR #527, opened 2022-03-19
remove leading zeros on receive

### LINKED ISSUE #499: Remove trailing zeroes from the auto-receive notification
image

both notifications should display "0.0022 XNO"

a similar code is already used in the `amount-split` pipe but it's worth mentioning that it needs to replace zeroes specifically in the factional part, unlike the regex below which would display 1500 XNO as 15 XNO

```js
const fractionalAmount = splitAmount.replace(/0+$/g, '');
```

\-

code for auto-receive notifications can be found in `wallet.service.ts` under `async processPendingBlocks()` method

geommr:
fixes #499


BitDesert:
Thanks @georgem3 !

PR #528, opened 2022-03-20
fix incorrectly styled buttons

### LINKED ISSUE #437: Incorrectly styled buttons on some layouts
good example (small viewport):

image

good example (large viewport):

image

(must be right-aligned)

code for those can be found in send.component.html around line 100

\-

bad examples:

1. address book new entry

image

2. rep book new entry

image

3. address book import confirmation

image

4. representatives page

image

5. sign/process block

image

geommr:
fixes #437


BitDesert:
Thanks for your contribution @georgem3 !

PR #529, opened 2022-03-21
fix Inaccurate description when no wallet is configured

### LINKED ISSUE #486: Inaccurate description when no wallet is configured
when changing Wallet Storage to None:

image

perhaps we could hide last 2 lines and only show this WHEN NO WALLET IS CONFIGURED:

image

geommr:
fixes #486

### DIFF
--- a/src/app/components/configure-app/configure-app.component.ts
+++ b/src/app/components/configure-app/configure-app.component.ts
@@ -348,7 +348,13 @@ export class ConfigureAppComponent implements OnInit {
     if (resaveWallet && newStorage === this.storageOptions[1].value) {
       const UIkit = window['UIkit'];
       try {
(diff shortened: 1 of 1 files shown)


keerifox:
thank you for contributing!

if i'm reading this correctly, the opening tag for bold text `<b style="font-size: 18px;">` should only be used if the wallet is configured (otherwise it's never closed, but also it shouldn't really be opened if there's no text. one of the line breaks `<br>` shouldn't be there either)

and maybe there could be just one check `this.walletService.isConfigured()`, even if it's simple logic it's probably not worth calling it once for every line

geommr:
should be fixed now

keerifox (approved):

BitDesert:
Thanks @georgem3 !

PR #530, opened 2022-03-23
use routes on wallet config

### LINKED ISSUE #447: Wallet import/creation should use more url routes
right now all the routes use https://nault.cc/configure-wallet which doesn't allow navigation via browser history

the proposal is to use the following:
https://nault.cc/configure-wallet (selection of options)
https://nault.cc/configure-wallet/new-wallet
https://nault.cc/configure-wallet/use-ledger
https://nault.cc/configure-wallet/import-seed
https://nault.cc/configure-wallet/import-mnemonic (selection of mnemonic options)
https://nault.cc/configure-wallet/import-nano-mnemonic
https://nault.cc/configure-wallet/import-bip39-mnemonic
https://nault.cc/configure-wallet/import-file
https://nault.cc/configure-wallet/import-private-key
https://nault.cc/configure-wallet/import-expanded-private-key

this would not only allow native browser navigation but also allow hotlinking to those pages

geommr:
fixes #447

### DIFF
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -30,6 +30,7 @@ const routes: Routes = [
   { path: 'account/:account', component: AccountDetailsComponent },
   { path: 'address-book', component: AddressBookComponent },
   { path: 'configure-wallet', component: ConfigureWalletComponent },
+  { path: 'configure-wallet/:mode', component: ConfigureWalletComponent },
   { path: 'configure-app', component: ConfigureAppComponent },
   { path: 'import-address-book', component: ImportAddressBookComponent },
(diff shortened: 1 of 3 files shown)


keerifox:
if a wallet is already configured, navigation to `/configure-wallet/new-wallet` should probably occur _after_ pressing OK in this dialogue window

image

at the moment if you press CANCEL, the url is already set to `/configure-wallet/new-wallet` and the button to create new wallet stops working

\-

another thing to test is visiting `/configure-wallet/new-wallet` directly.

it'd be nice if it displayed a confirmation popup. right now it does that but the translation strings are not initialized:

image

\-

everything else seems good to me!

thank you for the PR

PR #535, opened 2022-03-31
auto hide modals on 'back' navigation

### LINKED ISSUE #446: Pressing "Back" should close the visible modal
Right now pressing "Back" goes back in navigation history, but if there's a modal open, that "Back" action should close the modal instead (no URL must change)

This is particularly important to account/transaction menus on the compact account detail pages, but could apply to every modal.

geommr:
fixes #446


keerifox:
sorry only noticed this PR a short time ago,

does this prevent navigation or only hides the modals? per issue #446:

> Right now pressing "Back" goes back in navigation history, but if there's a modal open, that "Back" action should close the modal instead (no URL must change)

geommr:
My bad I must have missed this. The second commit should prevent navigation now

keerifox:
hmm i wonder if there's a cleaner way to solve it

the modal might become closed via numerous ways (clicking off-screen, proceeding with action etc.) and even if we listen to modal events ( https://getuikit.com/docs/modal#javascript ) to catch 90% of them, it feels like we may still have undesired effects by adding fake history state

PR #540, opened 2022-05-01
fix the max button that won't allow to send fractions of PAW

kduytsch:
Currently the Max button for sending $PAW  only allows to send entire value of PAW. This is handled by this function (in src/app/components/send/send.component.ts) : 

```
setMaxAmount() {
    const walletAccount = this.walletService.wallet.accounts.find(a => a.id === this.fromAccountID);
    if (!walletAccount) {
      return;
    }

    this.amountExtraRaw = walletAccount.balanceRaw;

    const nanoVal = this.util.nano.rawToNano(walletAccount.balance).floor();
    const maxAmount = this.getAmountValueFromBase(this.util.nano.nanoToRaw(nanoVal));
    this.amount = maxAmount.toNumber();
    this.syncFiatPrice();
  }
```

As you can read, we first get the balance, convert from RAW to PAW (well nano but we all understand it is a lazy fork) and floor it (so only rounded-low value, then convert back in RAW

To get decimal value we could use the following code instead 
```
  setMaxAmount() {
    const walletAccount = this.walletService.wallet.accounts.find(a => a.id === this.fromAccountID);
    if (!walletAccount) {
      return;
    }

    this.amountExtraRaw = walletAccount.balanceRaw;

    const maxAmount = this.getAmountValueFromBase(walletAccount.balance);
    this.amount = maxAmount.toNumber();
    this.syncFiatPrice();
  }
```


BitDesert:
I don't think this PR is correct. The PAW fork of Nault probably shouldn't be merged into the main repo.

Is this a proposed changed you want to include into Nault? Then create a PR with only this change here. Otherwise you can propose a change at the PAW fork repo.

keerifox:
> Is this a proposed changed you want to include into Nault? Then create a PR with only this change here.

even then the PR doesn't make sense because we specify additional raw in a separate variable, and it would break that behavior

kduytsch:
Well, to be honest it was my first PR ever

 

Now, that change is intended to be for PAW (Biome) only, but when on GITHUB or Biome and clicking on PR, I end-up on nault...

 

So not sure how to do it properly for Biome only (even if that change would be beneficial also on nano main branch)

 

BR

 

Karl

PR #573, opened 2023-01-22
fix BigNumber usage with undefined - assume 0 in that case.

Imp3:
Nano Node v24.0 returns no values for unused accounts rather than 0 in pre v24.0
Nano Node PR: https://github.com/nanocurrency/nano-node/pull/3790

### DIFF
--- a/src/app/services/wallet.service.ts
+++ b/src/app/services/wallet.service.ts
@@ -747,8 +747,8 @@ export class WalletService {
 
       if (!walletAccount) continue;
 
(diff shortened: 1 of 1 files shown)


keerifox (approved):

jamescoxon:
Is this still needed?

keerifox:
optional as it was fixed in v25 node release - https://docs.nano.org/releases/release-v25-0/#rpc-updates

jamescoxon:
I think we'll merge it as makes sense to assume 0 in case of issues in the future

Issue #414, opened 2021-07-06
Add OpenAlias Support


ppfeister:
**Is your feature request related to a problem? Please describe.**
Long hexadecimal wallet addresses are hard to remember, and they are _very_ unfriendly to the average user. This presents a huge barrier to wider adoption. Other currencies have tried to resolve this through new blockchain-based TLDs (ENS Domains, Unstoppable Domains, etc.), but those all rely on smart contracts - a feature Nano lacks. These new TLDs are also not supported by ICANN, meaning none of the standard DNS providers will actually resolve them.

**Describe the solution you'd like**
[OpenAlias](https://openalias.org/) is the perfect solution, at least until we find something better. The standard calls for a TXT record to be defined on a domain using any of the traditional TLDs (like .com). DNSSEC is encouraged by the standard to better prevent MITM attacks, but it is up to the individual implementation to require it or not. Formatted as following...
```
"oa1:nano recipient_address=nano_1111111111111111111111111111111111111111111111111111hifc8npp; recipient_name=John Doe;"
```
Once this record is defined, a user should be able to simply send Nano to example.com, instead of nano_111...npp. I encourage you to read the full write up on OpenAlias's homepage (https://openalias.org/).

**Describe alternatives you've considered**
The traditional solutions (i.e. decentralized name servers) either won't work with Nano, or would be very impractical. See first paragraph.

**Additional context**
No additional context at this time.

Joohansson:
It's a nice idea but I'm not convinced this can be made secure enough considered Nault is a client-side application and DNS lookup would use whatever DNS that the computer is using (I think). Say you get something malicious in the machine that intercepts any DNS lookup and returns a fake address. How to prevent that?

ppfeister:
That's definitely a concern, but I don't believe it would be an issue if implemented correctly. Standard DNS is definitely vulnerable to man-in-the-middle attacks (as you mentioned), but if DNSSEC or DoT (DNS over TLS) is used, it _should_ mitigate that.

After all, TLS is what we use for just about all of our other internet traffic, and it's proven very secure on that front. A lot of web browsers are starting to ship with DoT enabled by default for normal web traffic, as well (which is a good thing).

---
I'd like to see this enhancement use DoT by default, but I don't know how feasible that is just yet for Nault (I'd like to look more into that eventually). When it comes to DNSSEC, Nault has two possible solutions -
( a ) Outright require DNSSEC, and don't allow transactions to domains without it properly configured, or
[...]

PR #586, opened 2023-04-27
Nano.to Username Support

nano2dev:
Nano.to is a humble Nano public registry. 

The dataset is publicly via Github Pages at: https://nano.to/known.json

Anyone may register a Username for a small Nano price. 

I want to make a case with a really clean 1 Commit PR.

### DIFF
--- a/src/app/components/send/send.component.ts
+++ b/src/app/components/send/send.component.ts
@@ -212,7 +212,7 @@ export class SendComponent implements OnInit {
     
     addressBook
       .filter(a => a.name.toLowerCase().indexOf(search.toLowerCase()) !== -1).slice(0, 5)
-      .map(a => matches.push({ name: a.name + ' (Local Account)' }));
+      .map(a => matches.push({ name: a.name + ' (Local Account)', account: a.account }));
 
     this.known
       .filter(a => a.name.toLowerCase().indexOf(search.toLowerCase()) !== -1).slice(0, 5)
--- a/src/app/components/send/send.component.ts
+++ b/src/app/components/send/send.component.ts
@@ -14,6 +14,7 @@ import {NanoBlockService} from '../../services/nano-block.service';
 import { QrModalService } from '../../services/qr-modal.service';
 import { environment } from 'environments/environment';
 import { TranslocoService } from '@ngneat/transloco';
+import { HttpClient } from '@angular/common/http';
 
 const nacl = window['nacl'];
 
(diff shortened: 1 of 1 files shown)


keerifox:
per issue #414, we would like to have support for decentralized aliases that aren't conditional to one centralized entity providing the api or the names themselves

if we ever have that, the distinction with the local address book names must be abundantly clear to prevent mistakes similar to that time when someone tried to send a large sum of funds to John in their address book and they went to John on nanoto instead

nano2dev:
@keerifox You're always on time, to kill the party.

Nano, and unfortunately Nault faces a fundamental issue. 

No existing "embeddable" "Address Book". 

Nano.to is on Github. There's different API URLS with different caching strategies:

- https://nano.to/known.json
   - Github Pages (Changes Username update happens)
-  https://nano.to/2023/known.json 
   - (Immutable, "Github Release as JSON API")

For applications that demand, more immutability, there's an endpoint for that. 

We can choose to improve Nano with versioned controlled WEB2 into better Marketcap, or we can choose the "All Decentralized All The Time" route.

Nano does the currency thing well. **Nano.to** does the Username thing well.

vinibarbosabr:
IMHO there is a valid argument to be made that relying on a centralized database might create single points of failure that could be corrupted and are less secure than decentralized solutions.
If any decentralized solution is ever possible for a 'universal nano address book', it would be wonderful.

But in the meantime, I can also see a lot of value in what Nano.to is capable of doing in a centralized, but open-source manner. Users can get verified, and it's possible to tag 'nano.to aliases' so 'sender' can have more information before confirming the transaction.

Both of you are doing a great job and both of your points are valid.
Maybe there is a common-point to be found in-between. Maybe not.
I also respect the position of keeping everything the most decentralized possible, at the same time I recognize the huge challenges and possible set-backs it offers.

keerifox:
PR #587 allows use of aliases e.g. `@nano.to` (domain owner) or `@gr0vity@nano.to`, as long as nano.to adheres to the spec

nano2dev:
> PR #587 allows use of aliases e.g. `@nano.to` (domain owner) or `@gr0vity@nano.to`, as long as nano.to adheres to the spec

Native Nano.to implementation, albeit centralized is ~10 lines of code, provides better IP privacy, requires no warning popup, and requires no self-hosting.

Just saying ;)

It will take a bit, for Nano.to to implement this. Being a Static Web App brings querying e.g ```?names=Moon``` challenges.

Cloudflare Pages is so good, I *might* be able to pull this off: ```/.well-known/nano-currency.json?names=Moon```

[Nault.pro](https://Nault.pro) is a cool fork.

PR #589, opened 2023-05-17
Fix mismatching percentages in representative health tooltip

keerifox:
Text error introduced in #457 (transloco localization)

Source for 2% and 3% being in use in the code:

https://github.com/Nault/Nault/blob/cd6d388e60ce84affaa813991445734cdf64c49f/src/app/services/representative.service.ts#L163-L170


keerifox:
example of the issue

image

Issue #584, opened 2023-04-03
Desktop version fails to build ('openssl_fips' is not defined)


keerifox:
the error occurs while trying to build `abandonware/bluetooth-hci-socket@0.5.3-8` from source:

`npm ERR! gyp: name 'openssl_fips' is not defined while evaluating condition 'openssl_fips != ""' in binding.gyp while trying to load binding.gyp`

relevant node-gyp issue thread:
https://github.com/nodejs/node-gyp/issues/2673

dependency tree:
```shell
npm ls @abandonware/bluetooth-hci-socket@0.5.3-8
nault@1.18.0
`-- @ledgerhq/hw-transport-node-ble@6.27.1
  `-- @abandonware/noble@1.9.2-15
    `-- @abandonware/bluetooth-hci-socket@0.5.3-8
```

nanogarden:
When manually going through these steps, this error appears when running the following command:

`npm run desktop:local`

Appending the flag to that specific command fixes the issue:

`npm run desktop:local --openssl_fips=''`

nanogarden:
After some testing building in my computer (OS: ArchLinux, npm version: 8.19.4, node version: 16.20.1), my recommendation that would allow for an easy installation is to modify the `desktop:build` and `desktop:local` scripts directly in the packages.json file. 

Each one of these should include an openssl configuration flag.

These work:

`"desktop:build": "NODE_OPTIONS=--openssl-legacy-provider npm run wallet:build-desktop && npm run desktop:compile",`

`"desktop:local": "npm_config_openssl_fips='' electron-builder -p never"`

With this modification, the github build release should also work.

From looking at the above-cited issue (https://github.com/nodejs/node-gyp/issues/2673#issuecomment-1477682343), these flags may not be necessary with more recent versions of electron and electron-builder, so it may be a worthy project to update the dependencies.

nanogarden:
You can see the successful action run here, modifying only the packages.json file: https://github.com/nanogarden/Nault/actions/runs/5843583237

The "cross-env" npm package allows you to pass a variable in a way that will be accepted across platforms. So, the flag is passed in this way:

`"desktop:local": "cross-env npm_config_openssl_fips='' electron-builder -p never"`

And cross-env is added to the dependencies as:

`"cross-env": "^7.0.3",`

nanogarden:
[...]

PR #595, opened 2023-08-12
Potential temporary fix for desktop version failing to build

keerifox:
Likely not required on newer versions of electron.

Related to issue #584 - https://github.com/Nault/Nault/issues/584#issuecomment-1675918161

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+node-options="--openssl-legacy-provider"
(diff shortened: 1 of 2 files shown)


keerifox:
ubuntu build gets a checkmark now

windows gets `'npm_config_openssl_fips' is not recognized as an internal or external command`

mac unknown as it was aborted

PR #630, opened 2024-09-22
Add SpyNano node

BlickM:
Adds SpyNano.org's RPC and websocket to nodes list


jamescoxon (changes requested):
could you change 'shouldRandom' to 'false' for now, just while we get you established, we can then to the random list in the future. Thanks

jamescoxon (approved):

PR #657, opened 2024-11-30
Update FUNDING.yml

jamescoxon:
Removed mynano.ninja as doesn't now exist. Unsure if there is anything to replace it with.


keerifox:
> Unsure if there is anything to replace it with.

Could use nault sidebar's donation button's url, unless that's too meta

